PR #2, opened 2022-12-14
Beautify README

AShifter:
this is a critical patch
TIL you could use html tags in markdown and they just work???

Signed-off-by: Logan Lowe <AShifter@users.noreply.github.com>


jvyden (approved):
Weird way of doing it, it'll probably break in IDEs but it works lol

LGTM

PR #6, opened 2022-12-15
Add logger to `RequestContext`

Beyley:


Refresh.HttpServer/RequestContext.cs
```diff
@@ -1,8 +1,10 @@
 using System.Net;
+using NotEnoughLogs;
 
 namespace Refresh.HttpServer;
 
 public struct RequestContext
 {
     public HttpListenerRequest Request;
+    public LoggerContainer<RefreshContext>? Logger;
```
jvyden, line 9:
```suggestion
    public LoggerContainer<RefreshContext> Logger;
```

Shouldn't be nullable

jvyden (changes requested):

jvyden (approved):
i love my puter

PR #11, opened 2022-12-28
Dont create so many database contexts, and dont create database contexts if not needed

Beyley:
what it says on the tin


jvyden (approved):
Optimization is pog

Issue #19, opened 2023-03-17
Extend Tokens


jvyden:
- [x] Track Platform and GameVersion (this should also track if the token is an API token aswell) (#37)
- [x] Login date (https://github.com/LittleBigRefresh/Refresh/commit/28661735270b2d5602a4f660dc3c0293d8fbd228)
- [x] Add support for passing in a token in Bunkum (https://github.com/LittleBigRefresh/Bunkum/commit/d7a4617647540084b2b69c7106d1a279ede1713e)
- [ ] Maybe point to recent events? Could be used to display more information to user through API if a level upload was rejected for example. Would also be good for audit log

jvyden:
Recent events will probably use notifications (#27), closing in favor and as complete.

Issue #16, opened 2023-03-17
Recent Activity


jvyden:
- List of events in Realm
- Photos uploaded, news, new scores, etc.
- Should be exposed to API

Beyley:
https://github.com/LittleBigRefresh/Refresh/pull/23#discussion_r1141494186
Recent activity should include things like failed logins

jvyden:
This is now being worked on in this branch: https://github.com/LittleBigRefresh/Refresh/tree/recent-activity

jvyden:
Forgot to update status on this - this has been completed in #33.

Arcadius2006:
that was relatively fast

PR #23, opened 2023-03-17
Authentication via API

jvyden:
This is the next big step towards building out the layer that lets Refresh and `refresh-web` communicate.

# What is this?

This PR adds support for users to authenticate to the API. It also introduces a couple endpoints (`/api/v2/me`) and features (caching) that help the website side of things operate.

It also introduces a couple changes as I was working on Bunkum alongside this. Most importantly are the introduction of Middlewares.

# What is this *not*?

- This PR does not introduce authentication for game clients. Accounts will still be generated when a client logs in. This leads into the next point.
- This also isn't a way for users to sign up via the API. It is pure login.
- A full implementation of the API. There's definitely more work to be done!

These points above will absolutely be handled at a later date, but I want to get this merged as this PR is already getting quite large.

# Feedback

In particular, I'm looking for feedback with the password reset system and whether or not `Token`s should be separate for Game and API or joined as they are in this PR.

I'm also looking for feedback on how the API communicates back and forth with the client.

# Client

A reference client written in TypeScript can be found [here](https://github.com/LittleBigRefresh/refresh-web/tree/master/src/app/api), in `refresh-web`'s source code.

### DIFF
--- a/Refresh.GameServer/Authentication/Token.cs
+++ b/Refresh.GameServer/Authentication/Token.cs
@@ -19,12 +19,12 @@ public class Token : RealmObject
     
     // Realm can't store enums, use recommended workaround
     // ReSharper disable once InconsistentNaming (can't fix due to conflict with TokenType)
-    private string _tokenType { get; set; }
+    internal string _TokenType { get; set; }
 
     public TokenType TokenType
     {
-        get => Enum.Parse<TokenType>(this._tokenType);
-        set => this._tokenType = value.ToString();
+        get => Enum.Parse<TokenType>(this._TokenType);
+        set => this._TokenType = value.ToString();
     }
 
     public DateTimeOffset ExpiresAt { get; set; }
--- a/Refresh.GameServer/Authentication/Token.cs
+++ b/Refresh.GameServer/Authentication/Token.cs
@@ -8,13 +8,26 @@ namespace Refresh.GameServer.Authentication;
 
 #nullable disable
 
+[JsonObject(MemberSerialization.OptIn)]
 public class Token : RealmObject
 {
     [PrimaryKey]
     public ObjectId TokenId { get; set; } = ObjectId.GenerateNewId();
     
     // this shouldn't ever be serialized, but just in case let's ignore it
-    [JsonIgnore] [XmlIgnore] public string TokenData { get; set; }
+    [XmlIgnore] public string TokenData { get; set; }
     
+    // Realm can't store enums, use recommended workaround
+    // ReSharper disable once InconsistentNaming (can't fix due to conflict with TokenType)
+    private string _tokenType { get; set; }
+
+    public TokenType TokenType
+    {
+        get => Enum.Parse<TokenType>(this._tokenType);
+        set => this._tokenType = value.ToString();
+    }
+
+    public DateTimeOffset ExpiresAt { get; set; }
+
     public GameUser User { get; set; }
 }
--- a/Refresh.GameServer/Endpoints/Api/AuthenticationApiEndpoints.cs
+++ b/Refresh.GameServer/Endpoints/Api/AuthenticationApiEndpoints.cs
@@ -11,13 +11,28 @@ namespace Refresh.GameServer.Endpoints.Api;
 
 public class AuthenticationApiEndpoints : EndpointGroup
 {
+    // How many rounds to do for password hashing (BCrypt)
+    // 14 is ~1 second for logins and reset, which is fair because logins are a one-time thing
+    // 200 OK on POST '/api/v2/resetPassword' (1058ms)
+    // 200 OK on POST '/api/v2/auth' (1087ms)
+    //
+    // If increased, passwords will automatically be rehashed at login time to use the new WorkFactor
+    // If decreased, passwords will stay at higher WorkFactor until reset
+    private const int WorkFactor = 14;
+
     [ApiEndpoint("auth", Method.Post)]
     [Authentication(false)]
     public Response Authenticate(RequestContext context, RealmDatabaseContext database, ApiAuthenticationRequest body)
     {
         GameUser? user = database.GetUserByUsername(body.Username);
-        if (user == null) return new Response(HttpStatusCode.NotFound);
+        if (user == null)
+        {
+            return new Response(new ApiErrorResponse("The username or password was incorrect."), ContentType.Json, HttpStatusCode.Forbidden);
+        }
         
+        // database.SetUserPassword(user, null);
+        // return new Response("");
+
         // if this is a legacy user, have them create a password on login
         if (user.PasswordBcrypt == null)
         {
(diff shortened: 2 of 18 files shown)


Refresh.GameServer/Authentication/Token.cs
```diff
@@ -8,13 +8,26 @@ namespace Refresh.GameServer.Authentication;
...
     [PrimaryKey]
     public ObjectId TokenId { get; set; } = ObjectId.GenerateNewId();
     
     // this shouldn't ever be serialized, but just in case let's ignore it
-    [JsonIgnore] [XmlIgnore] public string TokenData { get; set; }
+    [XmlIgnore] public string TokenData { get; set; }
     
+    // Realm can't store enums, use recommended workaround
+    // ReSharper disable once InconsistentNaming (can't fix due to conflict with TokenType)
+    internal string _TokenType { get; set; }
```
Beyley, line 22: this should *probably* be an integer type, not a string, parsing an enum is slow

Refresh.GameServer/Endpoints/Api/AuthenticationApiEndpoints.cs
```diff
@@ -0,0 +1,134 @@
...
+
+    [ApiEndpoint("resetPassword", Method.Post)]
+    [Authentication(false)]
+    public Response ResetPassword(RequestContext context, RealmDatabaseContext database, ApiResetPasswordRequest body)
+    {
+        GameUser? user = database.GetUserFromResetTokenData(body.ResetToken);
+        if (user == null) return new Response(HttpStatusCode.Unauthorized);
+
+        if (body.PasswordSha512.Length != 128)
+            return new Response("Password is definitely not SHA512. Please hash the password - it'll work out better for both of us.",
```
Beyley, line 80: You should verify it is SHA512 using a regex here too, not just length

Refresh.GameServer/Extensions/MemoryStreamExtensions.cs
```diff
@@ -0,0 +1,8 @@
+using System.Text;
+
+namespace Refresh.GameServer.Extensions;
+
+internal static class MemoryStreamExtensions
+{
+    internal static void WriteString(this MemoryStream ms, string str) => ms.Write(Encoding.Default.GetBytes(str));
```
Beyley, line 7: `Encoding.Default` can vary from computer to computer, and [even from the same computer at different times](https://learn.microsoft.com/en-us/dotnet/api/system.text.encoding.default?view=net-8.0) 
I would reccomend manually specifying `UTF8` (since thats *probably* what you mean anyway)

Refresh.GameServer/Middlewares/DigestMiddleware.cs
```diff
@@ -0,0 +1,80 @@
...
+        using SHA1 sha = SHA1.Create();
+        string digestResponse = Convert.ToHexString(sha.ComputeHash(ms)).ToLower();
+
+        return digestResponse;
+    }
+    
+    // Referenced from Project Lighthouse
+    // https://github.com/LBPUnion/ProjectLighthouse/blob/d16132f67f82555ef636c0dabab5aabf36f57556/ProjectLighthouse.Servers.GameServer/Middlewares/DigestMiddleware.cs
+    // https://github.com/LBPUnion/ProjectLighthouse/blob/19ea44e0e2ff5f2ebae8d9dfbaf0f979720bd7d9/ProjectLighthouse/Helpers/CryptoHelper.cs#L35
+    // TODO: make this non-lbp specific, or implement middlewares and move to game server
```
Beyley, line 43: this `TODO` comment seems to not apply anymore

Refresh.GameServer/Middlewares/DigestMiddleware.cs
```diff
@@ -0,0 +1,80 @@
...
+    
+        string digestResponse = this.CalculateDigest(url, context.InputStream, auth);
+    
+        string digestHeader = !url.StartsWith("/lbp/upload/") ? "X-Digest-A" : "X-Digest-B";
+        string clientDigest = context.RequestHeaders[digestHeader] ?? string.Empty;
+        
+        context.ResponseHeaders["X-Digest-B"] = digestResponse;
+        if (clientDigest == digestResponse) return true;
+        
+        // this._logger.LogWarning(BunkumContext.Digest, $"Digest failed: {clientDigest} != {digestResponse}");
```
Beyley, line 57: we should log these to help people with lbp3 (or was it lbp2 that complains with wrong digest key), perhaps we should even save failed digest requests and display them to the user on the website to say "hey you patched incorrectly!" but this is more a "feature request" than a problem

Refresh.GameServer/Startup.cs
```diff
@@ -16,7 +18,7 @@
 BunkumHttpServer server = new()
 {
     AssumeAuthenticationRequired = true,
-    UseDigestSystem = true,
+    // UseDigestSystem = true,
```
Beyley, line 21: is there a reason this is just commented out and not removed outright?

Beyley (changes requested):

Refresh.GameServer/Endpoints/Api/AuthenticationApiEndpoints.cs
jvyden: I was thinking regex would be overkill for this check, but realistically it should be okay considering it's only a couple hundred more nanoseconds on a whole second

Will be fixed next commit

Refresh.GameServer/Extensions/MemoryStreamExtensions.cs
jvyden: Really? I've had issues with UTF8 encoding dropping binary characters as 0xEFBDBF...

I'm not exactly sure how this is possible given that the same docs you linked mention returning `UTF8Encoding`.

I'll test this out in a bit

Refresh.GameServer/Startup.cs
jvyden: Yeah, just a fragment from the process of updating Bunkum. This option no longer exists so I'll remove it.

Refresh.GameServer/Middlewares/DigestMiddleware.cs
jvyden: We can probably bundle this with Recent Activity / Tokens (tracked @ https://github.com/LittleBigRefresh/Refresh/issues/16 and https://github.com/LittleBigRefresh/Refresh/issues/19 respectively) and make that system a sort of "event log" that a user can view.

But yeah this isn't relevant to the PR so I'll resolve this lol
Feel free to note this idea in one of those issues

Refresh.GameServer/Extensions/MemoryStreamExtensions.cs
jvyden: Worked just fine, whatever issue I was having years ago was probably solved

Refresh.GameServer/Endpoints/Api/AuthenticationApiEndpoints.cs
```diff
@@ -0,0 +1,138 @@
...
+    [Authentication(false)]
+    public Response Authenticate(RequestContext context, RealmDatabaseContext database, ApiAuthenticationRequest body)
+    {
+        GameUser? user = database.GetUserByUsername(body.Username);
+        if (user == null)
+        {
+            return new Response(new ApiErrorResponse("The username or password was incorrect."), ContentType.Json, HttpStatusCode.Forbidden);
+        }
+        
+        // database.SetUserPassword(user, null);
```
Beyley, line 37: leaving undocumented commented code here is probably a bad idea

Beyley (changes requested):
one small thing, then it should b good to merge

Refresh.GameServer/Endpoints/Api/AuthenticationApiEndpoints.cs
jvyden: This was debugging code to quickly reset a user to a legacy user state to test password resets

I've removed it :+1:

jvyden:
Just tested everything fully, we are good to go so merging

PR #33, opened 2023-03-28
Recent activity

jvyden:
This PR implements Recent Activity.

This is implemented for both the API and in-game.

API preview:
image

In-game preview:
image

It's a little busted in-game, some events don't show properly, pagination is sketch, but this can all be fixed later. I'd prefer to get this out now and start filling the recent activity log asap even if the way to view it is scuffed.
While there are [many events](https://littlebigrefresh.github.io/Docs/recent-activity-events.txt) present in LBP, this PR only implements support for the following (as we haven't implemented everything yet):

- LevelUpload
- LevelFavourite
- UserFavourite

The "undo" counterparts of these events are unimplemented intentionally (as they do not show up in-game)

Pagination should work properly, but is untested

I tried to keep this as concise as possible, but this is a large & complicated feature in nature so it ended up being a fairly large PR (840 lines at the time of writing).


daggintosh:
i am proud, jvy den

Refresh.Analyzers/ActivityGenerator.cs
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    public void Execute(GeneratorExecutionContext context)
+    {
+        // while (!Debugger.IsAttached)
+        // {
+        //     Thread.Sleep(25);
+        // }
+        //
+        // Debugger.Break();
```
Beyley, line 151: odd test code probably shouldnt still be here (also pro tip: save the output to a file [Silk.NET does this and its very helpful], dont attach a debugger during this step to see the generated output)

Refresh.GameServer/Types/Activity/EventType.cs
```diff
@@ -0,0 +1,50 @@
...
+    [XmlEnum("review_level")]
+    LevelReview = 15,
+    // [XmlEnum("comment_on_user")]
+    // PostUserComment = 16,
+    // [XmlEnum("create_playlist")]
+    // PlaylistCreate = 17,
+    // [XmlEnum("heart_playlist")]
+    // PlaylistFavourite = 18,
+    // [XmlEnum("add_level_to_playlist")]
+    // PlaylistAddLevel = 19,
```
Beyley, line 49: why are a bunch of these commented out?

Beyley (changes requested):
only a few small things

Refresh.GameServer/Database/RealmDatabaseContext.Relations.cs
```diff
@@ -101,6 +105,8 @@ public bool UnfavouriteUser(GameUser userToFavourite, GameUser userFavouriting)
         {
             this._realm.Remove(relation);
         });
+        
+        // this.CreateUserUnfavouriteEvent(userFavouriting, userToFavourite);
```
Beyley, line 109: broken code should be documented as such

Refresh.GameServer/Types/Activity/EventType.cs
jvyden: These are for when these events are implemented in the future - uncommenting some will cause compilation errors (the source generator tries to implement types that don't yet exist, for example PhotoUpload will cause GamePhoto to be used)

Others are simply irrelevant, for example LevelUnpublish. I'm not sure if I'll end up implementing it in the future or not.

It's also to make sure people know what ids are already being used/potentially being used incase extra events are added in the future.

So I think it's okay to leave these here.

Refresh.Analyzers/ActivityGenerator.cs
jvyden: Yeah, I wasn't looking for output (I can view it in rider), rather I was looking for actual information about logic. Will clean up in next commit

Refresh.GameServer/Database/RealmDatabaseContext.Relations.cs
jvyden: Just gonna strip it entirely - I'll add it back if it does ever get used

Beyley (approved):

Beyley:
if it passes CI then lgtm

PR #34, opened 2023-03-29
Website: Ability to host from Refresh, bundle with CI builds

jvyden:
CI now automatically downloads the latest artifact from [refresh-web](https://github.com/LittleBigRefresh/refresh-web), and packages it with the artifact it creates when building Refresh.

This also rewrites the Release workflow, ensuring building is only done once to keep changes to building behaviour consistent. For example, I can foresee dotnet.yml being updated without release.yml being updated causing problems when a tag is pushed. This almost happened while making this PR.

### DIFF
--- a/.github/workflows/dotnet.yml
+++ b/.github/workflows/dotnet.yml
@@ -5,7 +5,6 @@ name: Build and Test
 
 on:
   push:
-    branches: [ "main" ]
   pull_request:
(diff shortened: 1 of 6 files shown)


jvyden:
Tested the whole workflow in a fork, tested API and everything, tried to exploit the middleware, it all worked great, so I'm merging

PR #43, opened 2023-05-07
Switch to using Realm's source generators

jvyden:
As Realm plans to deprecate the old "Fody" system of weaving code, I decided to move Refresh to the new system.

**This pull request needs to be tested thoroughly. Create a database on latest `main` branch if you haven't already, back it up, and then check if all data remains after migrating to this branch.**

Development of this caused an accidental partial wipe of data in my dev environment, which is why I ask that you be careful. I think it'll be fine, because the issue was I lacked proper migrations due to changes that the new system made. I've written those migrations so everything should be okay.


jvyden:
Had @Beyley test on [Discord](https://discord.com/channels/1049223665243389953/1049225857350254632/1105632932560896000), everything in the database stayed as expected. I also did some of my own testing and it worked aswell. I feel confident doing this, so, merging!

jvyden:
..well, I would if the build passed.

PR #92, opened 2023-08-13
Add docker support

Slendy:
## How to run
```
docker compose up
```
Data is stored in the `data` directory at the root of the git repository. 

The refresh process is run as non-root and is process id 1 inside the container
```
root@5a2c90c09b73:/refresh# ps aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
refresh        1  2.8  0.3 272237924 83968 ?     Ssl  00:21   0:01 /refresh/app/Refresh.GameServer
```

Nuget restores are cached unless any of the .sln or .csproj files are changed


sudokoko (approved):
lgtm- wait, i'm on the wrong repo :trollface:

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+WORKDIR /build
+
+COPY *.sln ./
+COPY **/*.csproj ./
+
+RUN dotnet sln list | grep ".csproj" \
+    | while read -r line; do \
+    mkdir -p $(dirname $line); \
+    mv $(basename $line) $(dirname $line); \
+    done;
```
jvyden, line 12: What does this do?

Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+# Final running container
+
+FROM mcr.microsoft.com/dotnet/runtime:7.0.10-bookworm-slim AS final
+
+# Add non-root user
+RUN set -eux && \
+apt update && \
+apt install -y gosu && \
+rm -rf /var/lib/apt/lists/* && \
+gosu nobody true && \
```
jvyden, line 29: Most of this I can follow, but I don't understand this invocation of gosu

jvyden (changes requested):
Haven't tested locally, that will happen soon after this review but the scripts do need some further explanation.

We *would* like support for Docker (as it follows our ease of use practices), even if we don't have any plans to use it in production, but some more information & on how exactly these scripts function would be helpful

Dockerfile
Slendy: It’s checking that the gosu binary works by running /bin/true as the nobody user. Gosu is used in many docker entrypoint scripts rather than sudo because sudo can cause weird TTY and signal forwarding behavior

Dockerfile
Slendy: The wildcard copy command before this doesn’t keep the original directory structure of the .csproj files. This just restores the original directory structure. Refresh.GameServer.csproj will get moved to Refresh.GameServer/Refresh.GameServer.csproj. This step happens before copying any actual code because in the docker build cache this step doesn’t have to be rerun unless any of the .csproj files change. If you copied the entire repo first then any change would trigger a cache invalidation. You could alternatively copy the project files manually and specify the folder they should be in.

jvyden:
Does the invocation of `docker compose up` cause a rebuild of the docker image? I'm writing some documentation for the README and wondering if you can just run a `git checkout vX.X.X; docker compose up` to update.

sudokoko:
In my experience you have to use `docker compose up --build` which builds
and then (re)starts containers

jvyden:
Cool, thanks.

jvyden (approved):
Local testing went well, so I'm merging this. Thank you for your contribution!

PR #103, opened 2023-08-26
Beautify readme.md with some emojis

speedyman09:
Lmk if anything should be changed

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-# Refresh
+# Refresh :earth_americas:
 
-A second-generation custom server for LittleBigPlanet that focuses on ease of use and reliability.
+A second-generation custom server for LittleBigPlanet that focuses on ease of use and reliability. :earth_americas::video_game:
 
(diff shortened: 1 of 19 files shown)


jvyden (changes requested):
I think it's *okay* for the headers, but putting them mid-paragraph is a little much.

I probably won't merge this either way since emojis in repositories' readmes have always felt wrong to me, but I'd like to see what others think

speedyman09:
@turecross321 what is your opinion on this?

turecross321:
I think it looks a bit tacky with emojis mid-paragraph too, but I like them in headers.
That being said, I think more than one emoji per header is too much, and it might be worth experimenting with putting the emoji before the text as opposed to after it.

jvyden:
I mostly agree with Ture. By the way, we should probably use the actual Unicode characters - most markdown parsers don't support shortcodes.

speedyman09:
Can you explain a usecase where someone would be using a custom markdown parser?

jvyden:
image

PR #104, opened 2023-08-26
Fix README Typo

turecross321:


jvyden (approved):
Legalties isn't a word??? Damn, alright then

PR #139, opened 2023-08-28
Verify all the things + fix grief reports

Beyley:
This does some code cleanup and sanitizes user inputs a lot more (+ a fix for grief reports)

### DIFF
--- a/Refresh.GameServer/Endpoints/Game/DataTypes/Request/GameLevelRequest.cs
+++ b/Refresh.GameServer/Endpoints/Game/DataTypes/Request/GameLevelRequest.cs
@@ -1,0 +1,140 @@
+using System.Xml.Serialization;
+using Refresh.GameServer.Authentication;
+using Refresh.GameServer.Database;
+using Refresh.GameServer.Endpoints.ApiV3.DataTypes;
+using Refresh.GameServer.Endpoints.Game.DataTypes.Response;
+using Refresh.GameServer.Services;
+using Refresh.GameServer.Types;
+using Refresh.GameServer.Types.Levels;
+using Refresh.GameServer.Types.Levels.SkillRewards;
+using Refresh.GameServer.Types.Matching;
+using Refresh.GameServer.Types.Reviews;
+using Refresh.GameServer.Types.UserData;
+
+namespace Refresh.GameServer.Endpoints.Game.DataTypes.Request;
+
+[XmlRoot("slot")]
+[XmlType("slot")]
+public class GameLevelRequest : IDataConvertableFrom<GameLevelRequest, GameLevel>
+{
+    [XmlElement("id")] public required int LevelId { get; set; }
+
+    [XmlElement("name")] public required string Title { get; set; }
+    [XmlElement("icon")] public required string IconHash { get; set; }
+    [XmlElement("description")] public required string Description { get; set; }
+    [XmlElement("location")] public required GameLocation Location { get; set; }
+
+    [XmlElement("game")] public required int GameVersion { get; set; }
+    [XmlElement("rootLevel")] public required string RootResource { get; set; }
+
+    [XmlElement("firstPublished")] public required long PublishDate { get; set; } // unix seconds
+    [XmlElement("lastUpdated")] public required long UpdateDate { get; set; }
+    
+    [XmlElement("minPlayers")] public required int MinPlayers { get; set; }
+    [XmlElement("maxPlayers")] public required int MaxPlayers { get; set; }
+    [XmlElement("enforceMinMaxPlayers")] public required bool EnforceMinMaxPlayers { get; set; }
+    
+    [XmlElement("sameScreenGame")] public required bool SameScreenGame { get; set; }
+
+    [XmlAttribute("type")] public string Type { get; set; } = "user";
+
+    [XmlElement("npHandle")] public SerializedUserHandle Handle { get; set; } = null!;
+    
+    [XmlElement("heartCount")] public required int HeartCount { get; set; }
+    
+    [XmlElement("playCount")] public required int TotalPlayCount { get; set; }
+    [XmlElement("uniquePlayCount")] public required int UniquePlayCount { get; set; }
+
+    [XmlElement("yourDPadRating")] public int YourRating { get; set; }
+    [XmlElement("thumbsup")] public required int YayCount { get; set; }
+    [XmlElement("thumbsdown")] public required int BooCount { get; set; }
+    
+    [XmlArray("customRewards")]
+    [XmlArrayItem("customReward")]
+    public required List<GameSkillReward> SkillRewards { get; set; }
+
+    [XmlElement("mmpick")] public required bool TeamPicked { get; set; }
+    [XmlElement("resource")] public List<string> XmlResources { get; set; } = new();
+    [XmlElement("playerCount")] public int PlayerCount { get; set; }
+
+    public static GameLevelRequest? FromOldWithExtraData(GameLevel? old, GameDatabaseContext database, MatchService matchService, GameUser user)
+    {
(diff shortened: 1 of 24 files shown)


Refresh.GameServer/Endpoints/Game/DataTypes/Request/GameLevelRequest.cs
```diff
@@ -0,0 +1,140 @@
...
+    [XmlElement("npHandle")] public SerializedUserHandle Handle { get; set; } = null!;
+    
+    [XmlElement("heartCount")] public required int HeartCount { get; set; }
+    
+    [XmlElement("playCount")] public required int TotalPlayCount { get; set; }
+    [XmlElement("uniquePlayCount")] public required int UniquePlayCount { get; set; }
+
+    [XmlElement("yourDPadRating")] public int YourRating { get; set; }
+    [XmlElement("thumbsup")] public required int YayCount { get; set; }
+    [XmlElement("thumbsdown")] public required int BooCount { get; set; }
```
jvyden, line 50: These shouldn't be here for a request

Refresh.GameServer/Endpoints/Game/DataTypes/Request/GameLevelRequest.cs
```diff
@@ -0,0 +1,140 @@
...
+    
+    [XmlArray("customRewards")]
+    [XmlArrayItem("customReward")]
+    public required List<GameSkillReward> SkillRewards { get; set; }
+
+    [XmlElement("mmpick")] public required bool TeamPicked { get; set; }
+    [XmlElement("resource")] public List<string> XmlResources { get; set; } = new();
+    [XmlElement("playerCount")] public int PlayerCount { get; set; }
```
jvyden, line 58: Same with these

Refresh.GameServer/Endpoints/Game/DataTypes/Request/GameLevelRequest.cs
```diff
@@ -0,0 +1,140 @@
...
+
+        if (old.Publisher == null)
+        {
+            response.Type = "developer";
+        }
+        else
+        {
+            response.Type = "user";
+            response.Handle = SerializedUserHandle.FromUser(old.Publisher);
+        }
```
jvyden, line 106:
```suggestion
```

Refresh.GameServer/Endpoints/Game/DataTypes/Request/GameLevelRequest.cs
```diff
@@ -0,0 +1,140 @@
...
+        response.FillInExtraData(database, matchService, user);
+
+        return response;
+    }
+
+    public static GameLevelRequest? FromOld(GameLevel? old)
+    {
+        if (old == null) return null;
+
+        GameLevelRequest response = new()
```
jvyden, line 74:
```suggestion
        GameLevelRequest request = new()
```

Refresh.GameServer/Endpoints/Game/DataTypes/Request/GameLevelRequest.cs
```diff
@@ -0,0 +1,140 @@
...
+
+    private void FillInExtraData(GameDatabaseContext database, MatchService matchService, GameUser user)
+    {
+        GameLevel? level = database.GetLevelById(this.LevelId);
+        if (level == null) throw new InvalidOperationException("Cannot fill in level data for a level that does not exist.");
+        
+        this.YourRating = (int)database.GetRatingByUser(level, user);
+        this.PlayerCount = matchService.GetPlayerCountForLevel(RoomSlotType.Online, this.LevelId);
+    }
+}
```
jvyden, line 140: This doesn't need to be here since it doesn't originate from the database

Refresh.GameServer/Endpoints/Game/Levels/PublishEndpoints.cs
```diff
@@ -13,10 +15,57 @@ namespace Refresh.GameServer.Endpoints.Game.Levels;
...
+        if (body.BooCount != 0 || body.YayCount != 0 || body.HeartCount != 0)
+        {
+            logger.LogWarning(BunkumContext.UserContent, $"User {user.Username} attempted to force non-0 boo/yay/heart counts! This is very likely a forged request.");
+            return false;
+        }
+
+        if (body.PlayerCount > 4)
+        {
+            return false;
+        }
```
jvyden, line 57: Not sure if it's worth logging/checking these

Refresh.GameServer/Endpoints/Game/ResourceEndpoints.cs
```diff
@@ -32,6 +33,9 @@ public class ResourceEndpoints : EndpointGroup
         if (gameAsset == null)
             return BadRequest;
 
+        long earliestTime = new DateTimeOffset(2007, 1, 1, 0, 0, 0, TimeSpan.Zero).ToUnixTimeSeconds();
```
jvyden, line 36: This should probably be a constant/static variable somewhere

Maybe `EarliestDate` on `IDateTimeProvider`

Refresh.GameServer/Endpoints/Game/UserEndpoints.cs
```diff
@@ -104,6 +97,12 @@ public SerializedUserList GetMultipleUsers(RequestContext context, GameDatabaseC
             return null;
         }
 
+        if (data.Description is { Length: > 4096 })
+        {
+            database.AddErrorNotification("Profile update failed", "Your profile failed to update because the description was too long. (Max 4096)", user);
+            return null;
+        }
```
jvyden, line 104:
```suggestion
        const int maxDescriptionLength = 4096;
        if (data.Description is { Length: > maxDescriptionLength })
        {
            database.AddErrorNotification("Profile update failed", $"Your profile failed to update because the description was too long. The max length is {maxDescriptionLength}.", user);
            return null;
        }
```

RefreshTests.GameServer/GameServer/TestRefreshGameServer.cs
```diff
@@ -29,6 +31,8 @@ public override void Start()
         // this._workerManager.Start();
     }
 
+    protected override IDateTimeProvider GetTimeProvider() => new MockDateTimeProvider();
```
jvyden, line 34: Should be exposed to tests for controlling via a property

jvyden (changes requested):
Good start, but I don't like how level requests are done

jvyden:
I'm guessing tests broke due to missing NAT type in payload?

Beyley:
> I'm guessing tests broke due to missing NAT type in payload?

Yes, i have an unpushed commit fixing that, and also adding NAT matching tests

Refresh.GameServer/Endpoints/Game/Levels/PublishEndpoints.cs
Beyley: Yeah it may just be worth removing those fields from the GameLevelRequest object altogether, since they dont *need* to exist for requests, mostly they are there to catch any attempted request forgery that someone may try

jvyden: At the end of the day I think it's up to instance moderators to handle request forgery, but we absolutely should block it where we should. In this case, I don't like having these checks because:

- The forgery wouldn't work anyways due to the nature of how Refresh handles Levels
- It creates a mess in the Request class
- The Request class is meant to silently drop things like this in the first place

jvyden: Plus, having the extra properties could lead to confusion for developers

jvyden (approved):

PR #148, opened 2023-09-02
Add chat commands

Beyley:
Adds basic chat command to force who you match with


Refresh.GameServer/Endpoints/Game/UserEndpoints.cs
```diff
@@ -168,15 +169,35 @@ public SerializedUserList GetMultipleUsers(RequestContext context, GameDatabaseC
...
         Debug.Assert(user != null);
         Debug.Assert(body != null);
-        
-        context.Logger.LogInfo(BunkumContext.Filter, $"<{user}>: {body}");
 
         //TODO: Add filtering
         
+        if (commandService.IsPublishing(user.UserId))
+        {
+            context.Logger.LogInfo(BunkumContext.Filter, $"Publish filter {body}");
```
jvyden, line 181:
```suggestion
            context.Logger.LogInfo(BunkumContext.Filter, $"Publish filter '{body}'");
```

Refresh.GameServer/Services/CommandService.cs
```diff
@@ -0,0 +1,103 @@
...
+    private readonly HashSet<ObjectId> _usersPublishing = new();
+
+    /// <summary>
+    /// Start tracking the user, eg. they started publishing
+    /// </summary>
+    /// <param name="id">The user ID</param>
+    public void StartPublishing(ObjectId id)
+    {
+        //Unconditionally add the user to the set
+        _ = this._usersPublishing.Add(id);
```
jvyden, line 29: This isn't pure so we dont need a blank assignment

Refresh.GameServer/Services/CommandService.cs
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    /// <summary>
+    /// Stop tracking the user, eg. they stopped publishing
+    /// </summary>
+    /// <param name="id">The user ID</param>
+    public void StopPublishing(ObjectId id)
+    {
+        //Unconditionally remove the user from the set
+        _ = this._usersPublishing.Remove(id);
```
jvyden, line 39: Same here

Refresh.GameServer/Services/CommandService.cs
```diff
@@ -0,0 +1,103 @@
...
+    public bool IsPublishing(ObjectId id) => this._usersPublishing.Contains(id);
+
+    /// <summary>
+    /// Parse a command string into a command object
+    /// </summary>
+    /// <param name="str">Command string</param>
+    /// <returns>Parsed command</returns>
+    /// <exception cref="FormatException">When the command is in an invalid format</exception>
+    [Pure]
+    public Command ParseCommand(string str)
```
jvyden, line 51: I know we talked aboutthis last night, but the fact that this could definitely be using Spans for this function to make it allocation-less still irks me. I can do this for you if you'd like

Refresh.GameServer/Types/Commands/Command.cs
```diff
@@ -0,0 +1,3 @@
+namespace Refresh.GameServer.Types.Commands;
+
+public record Command(string Name, string? Arguments);
```
jvyden, line 3: Might be possible to use a ref struct here, we dont store commands persistently.

Also, the name could be improved. Something like CommandData? CommandInvocation?

jvyden (changes requested):
Looks good, just a couple questions tho

Refresh.GameServer/Services/CommandService.cs
Beyley: This was mostly done to be explicit about whats happening (ignoring the result) but the comment does that already, so i can remove the `_ =`

PR #165, opened 2023-09-23
Send allOk response for batch filtering

Beyley:
Not doing this caused some PSP levels to fail to upload, crashing the console


jvyden (approved):
yeah seems legit

PR #171, opened 2023-09-25
Lie to the PSP, and return OK instead of Unauthorized with favourite level requests

Beyley:


jvyden (approved):

jvyden:
i feel like linus torvalds rn
i dont commit i just review and merge

PR #174, opened 2023-09-27
Implement UsedSlots for LBP1/3/PSP

### LINKED ISSUE #173: Implement GameUserResponse UsedSlots
On PSP the game will always display `0/20`, see [this line](https://github.com/LittleBigRefresh/Refresh/blob/ba164c00c7e10ab544590545bc5864549a670f48/Refresh.GameServer/Endpoints/Game/DataTypes/Response/GameUserResponse.cs#L75) of code, which needs to be filled in

Beyley:
Closes #173

I believe we should be filtering the count for each game, eg. LBP2 levels shouldnt count towards the LBP1 count, and LBP1 shouldnt count towards the LBP PSP count, but this is fine for now, and at least fills in this count so *something* is displayed in-game

### DIFF
--- a/Refresh.GameServer/Endpoints/Game/DataTypes/Response/GameUserResponse.cs
+++ b/Refresh.GameServer/Endpoints/Game/DataTypes/Response/GameUserResponse.cs
@@ -72,16 +72,12 @@ public class GameUserResponse : IDataConvertableFrom<GameUserResponse, GameUser>
             EntitledSlots = 100,
             EntitledSlotsLBP2 = 100,
             EntitledSlotsLBP3 = 100,
-            UsedSlots = old.PublishedLevels.Count(),
-            UsedSlotsLBP2 = old.PublishedLevels.Count(),
-            UsedSlotsLBP3 = old.PublishedLevels.Count(),
+            UsedSlots = 0,
+            UsedSlotsLBP2 = 0,
+            UsedSlotsLBP3 = 0,
             PurchasedSlotsLBP2 = 0,
             PurchasedSlotsLBP3 = 0,
         };
-
-        response.FreeSlots = 100 - response.UsedSlots;
-        response.FreeSlotsLBP2 = 100 - response.UsedSlotsLBP2;
-        response.FreeSlotsLBP3 = 100 - response.UsedSlotsLBP3;
 
         return response;
     }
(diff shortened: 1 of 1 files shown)


jvyden:
No, that behavior is actually correct. All the counts are per-game - e.g. you can upload a total of 100 levels for LBP2, but you can still upload 100 more on LBP1

Beyley:
> No, that behavior is actually correct. All the counts are per-game - e.g. you can upload a total of 100 levels for LBP2, but you can still upload 100 more on LBP1

Yes, that is what we do not currently do, right now we tally counts for all games for all counts, meaning if you upload 100 LBP2 levels, LBP1 will say you have also uploaded 100 LBP1 levels

jvyden:
No, I truly do mean that they're separate - LBP1 isn't supposed to know/care about the 100 levels on LBP2

jvyden:
Oh, misunderstood.

jvyden:
Okay, now I'm caught up and I'm actually thinking about the words I say.

If that's the case I'd rather not merge this seeing as it doesn't really do much for the user. If anything it'll probably just break things worse incase you have 100 LBP2 levels uploaded. LBP1 will refuse to even let you attempt if your UsedSlots is full.

Implementing per-game level counts should be trivial so I don't see why we shouldn't. (especially given the fact that we have LINQ access right here even though that's generally considered bad practice in Refresh)

Beyley:
> Okay, now I'm caught up and I'm actually thinking about the words I say.
> 
> If that's the case I'd rather not merge this seeing as it doesn't really do much for the user. If anything it'll probably just break things worse incase you have 100 LBP2 levels uploaded. LBP1 will refuse to even let you attempt if your UsedSlots is full.
> 
> Implementing per-game level counts should be trivial so I don't see why we shouldn't. (especially given the fact that we have LINQ access right here even though that's generally considered bad practice in Refresh)

How would i go about separating LBP1 and LBP PSP levels? they use the same field name, would i need to pass the game version all the way down to this method?

jvyden:
You should already have this information, IIRC there was some previous work done for Vita that made serialized slots depend on the version

Beyley:
> You should already have this information, IIRC there was some previous work done for Vita that made serialized slots depend on the version

I see this now, would you happen to know if Vita uses the LBP2 or LBP1 FreeSlots count (or would that not matter, and for example if you are on LBP Vita, send the LBP Vita count on all fields)

jvyden:
Vita should use LBP2 slots.

Beyley:
Implemented it properly for all games

Refresh.GameServer/Endpoints/Game/DataTypes/Response/GameUserResponse.cs
```diff
@@ -103,5 +99,46 @@ private void FillInExtraData(GameUser old, TokenGame gameVersion)
...
+                this.FreeSlotsLBP2 = 100 - this.UsedSlotsLBP2;
+                //Fill out LBP1 levels
+                goto case TokenGame.LittleBigPlanet1;
+            }
+            case TokenGame.LittleBigPlanetVita: { 
+                //Match all LBP Vita levels
+                this.UsedSlotsLBP2 = old.PublishedLevels.Count(x => x._GameVersion == (int)TokenGame.LittleBigPlanetVita);
+                this.FreeSlotsLBP2 = 100 - this.UsedSlotsLBP2;
+                //Fill out LBP1 levels
+                goto case TokenGame.LittleBigPlanet1;
```
jvyden, line 125: Can probably skip this entirely, Vita is it's own little world

jvyden (changes requested):
One little nitpick

Refresh.GameServer/Endpoints/Game/DataTypes/Response/GameUserResponse.cs
Beyley: Ah, for some reason i remember being told Vita could load LBP1 levels, in that case i'll just `break` out

jvyden (approved):

PR #175, opened 2023-09-27
Update NUnit.Analyzers from 3.6.1 to 3.8.0

### LINKED ISSUE #163: Bump NUnit.Analyzers from 3.6.1 to 3.7.0
This package brings in some breaking changes that prevent compilation. Work should be fairly trivial.

https://github.com/LittleBigRefresh/Refresh/actions/runs/6218288715/job/16874547419

Beyley:
Closes #163

The weird extra stuff was added by rider, so im assuming whatever its doing is good, since the analyzers do seem to run


jvyden (approved):
Sweet

PR #195, opened 2023-10-07
Implement story leaderboards + LBP1 multiplayer leaderboards

Beyley:


jvyden (changes requested):
I have a couple concerns with how story leaderboards were implemented in this PR. Tying scores to a developer ID means we won't have a `GameLevel` associated with story levels and we won't be able to implement things like reviews/ratings/comments/stats for story levels in the future. The absence of a `GameLevel` association also means we can't discover scores for story levels via the API at the moment since that path requires a level ID.

Level authors in Refresh can be null, so it's better to just create new levels on the database as they are needed. Unfortunately we can't work around this by creating a fake `GameLevel` locally, as they'll be required in the future as we store comments and such things in the database.

I'm in favor of adding the `TokenGame` (and maybe even the `TokenPlatform`) to the score, but we can't have it be nullable since that complicates things both in Realm and for API users. As a okay-ish default you can set it to the level's game version during migration, even though it won't be 100% accurate given that levels are forwards compatible.

Beyley:
> I have a couple concerns with how story leaderboards were implemented in this PR. Tying scores to a developer ID means we won't have a `GameLevel` associated with story levels and we won't be able to implement things like reviews/ratings/comments/stats for story levels in the future. The absence of a `GameLevel` association also means we can't discover scores for story levels via the API at the moment since that path requires a level ID.
> 
> Level authors in Refresh can be null, so it's better to just create new levels on the database as they are needed. Unfortunately we can't work around this by creating a fake `GameLevel` locally, as they'll be required in the future as we store comments and such things in the database.
> 
> I'm in favor of adding the `TokenGame` (and maybe even the `TokenPlatform`) to the score, but we can't have it be nullable since that complicates things both in Realm and for API users. As a okay-ish default you can set it to the level's game version during migration, even though it won't be 100% accurate given that levels are forwards compatible.

So the process would be 
`database.GetStoryLevel(id, game)` and itd return (and create if necessary) a level which is associated with the developer id, then all the database/score actions would go off of that?

Refresh.GameServer/Database/GameDatabaseContext.Leaderboard.cs
```diff
@@ -29,30 +31,26 @@ public GameSubmittedScore SubmitScore(SerializedScore score, GameUser user, Game
...
     
-    [UsedImplicitly] private record ScoreLevelWithPlayer(GameLevel level, GameUser player);
+    [UsedImplicitly] private record ScoreLevelWithPlayer(GameLevel Level, GameUser Player);
 
     public DatabaseList<GameSubmittedScore> GetTopScoresForLevel(GameLevel level, int count, int skip, byte type, bool showDuplicates = false)
     {
-        IEnumerable<GameSubmittedScore> scores = this._realm.All<GameSubmittedScore>()
-            .Where(s => s.Level == level && s.ScoreType == type)
-            .OrderByDescending(s => s.Score)
+        IEnumerable<GameSubmittedScore> scores = this._realm.All<GameSubmittedScore>().Where(s => s.ScoreType == type && s.Level == level)
```
jvyden, line 38: Why are we removing the sorting here?

Refresh.GameServer/Database/GameDatabaseContext.Levels.cs
```diff
@@ -27,6 +28,32 @@ public void AddLevel(GameLevel level)
         });
     }
 
+    public GameLevel GetStoryLevelById(int id, TokenGame game)
```
jvyden, line 31: Developer ids aren't unique per game, we probably don't need to pass in `TokenGame` for retrieval

Refresh.GameServer/Types/Levels/GameLevelSource.cs
```diff
@@ -0,0 +1,7 @@
+namespace Refresh.GameServer.Types.Levels;
+
+public enum GameLevelSource
+{
+    User,
+    Story,
```
jvyden, line 6: Should add XMLDoc here

Refresh.GameServer/Database/GameDatabaseContext.Leaderboard.cs
Beyley: that was an accident, will re add

Refresh.GameServer/Database/GameDatabaseContext.Levels.cs
Beyley: PSP IDs start from 0, so i assume at least with LBP PSP, there would be clash with LBP1, assuming those start with `0` too (not sure)

jvyden: They don't, LBP1 should start somewhere in the thousands range

jvyden:
We should probably also limit the database functions to fetching only user levels since we're creating new levels in the database for developer levels. We should also introduce a dedicated category for developer slots for API use.

Beyley:
> We should probably also limit the database functions to fetching only user levels since we're creating new levels in the database for developer levels. We should also introduce a dedicated category for developer slots for API use.

I already limited the database functions to fetch only user levels, and added a `Source` property on `GameLevel` to tell whether its a `User` or `Story` level (although i dont think you are able to ask the API for info about that yet), although if categories are wanted instead i can do that

jvyden:
I don't foresee a case where the extra categories we have on user levels (ratings, hearts, plays, etc.) would be needed for developer levels so I think it's okay if we just have one category for all the developer levels.

We don't need to add the `source` property for the returned JSON since it'll be implied from the endpoint you got it from.

Beyley:
> I don't foresee a case where the extra categories we have on user levels (ratings, hearts, plays, etc.) would be needed for developer levels so I think it's okay if we just have one category for all the developer levels.
> 
> We don't need to add the `source` property for the returned JSON since it'll be implied from the endpoint you got it from.

Sorry, i completely misunderstood what you had meant at first, yes i will add an API category for developer levels

Beyley:
I added a simple developer level API category, how should this be handled on the site? Currently it just looks like this
image

jvyden:
Just hide this category for now

Beyley:
> Just hide this category for now, see #196

How would i go about hiding it? would this be done on the `refresh-web` side? I dont see a `hide` field on LevelCategory

jvyden:
Oh, right, this *would* be done in `refresh-web`. I'll make a commit to do that now.

Refresh.GameServer/Types/Levels/Categories/CategoryService.cs
```diff
@@ -69,6 +69,13 @@ public class CategoryService : EndpointService
             IconHash = "g820626",
             FontAwesomeIcon = "certificate",
         },
+        new LevelCategory("developerLevels", Array.Empty<string>(), false, nameof(GameDatabaseContext.GetDeveloperLevels))
+        {
+            Name = "Developer Levels",
+            Description = "Levels made by the developers of LBP",
+            IconHash = "g820626",
+            FontAwesomeIcon = "terminal",
+        },
```
jvyden, line 78:
```suggestion
        new LevelCategory("developer", Array.Empty<string>(), false, nameof(GameDatabaseContext.GetDeveloperLevels))
        {
            Name = "Story Levels",
            Description = "Levels bundled with LittleBigPlanet",
            IconHash = "g820604",
            FontAwesomeIcon = "certificate",
        },
```

jvyden:
Made the change in refresh-web

Beyley:
> Made the change in refresh-web

In that case this should be ready for merge, assuming no more review comments are found

jvyden (approved):
WCGW

PR #197, opened 2023-10-14
Implement count query parameter for play/user/{id}

### LINKED ISSUE #178: Batched level plays (ex: `/lbp/play/user/1?count=3`).

Beyley:
This is only used on PSP, so the logic is only enabled on PSP callers

Closes #178


jvyden (approved):
Oops, missed this one. I think there should likely be a limit but it's not too big of a deal since we primarily use unique plays rather than total plays.

PR #199, opened 2023-10-16
Add option to redirect grief reports to photos

Beyley:
This adds photo support to PSP, and allows you to take really quick photos (with UI) on the mainline games almost anywhere too

Marking as draft as [this](https://github.com/Beyley/Refresh/blob/59c21971bec14b7f59b85bf4c7a1ece933fe1054/Refresh.GameServer/Endpoints/Game/ReportingEndpoints.cs#L37) needs to be resolved, so we can remove [this hack](https://github.com/Beyley/Refresh/blob/98b7e77e1e0341b914746fd22515388370999855/Refresh.GameServer/Types/Photos/SerializedPhotoSubject.cs#L26), but im not sure what the range of values are supposed to be

### DIFF
--- a/Refresh.GameServer/Endpoints/ApiV3/DataTypes/Response/ApiExtendedGameUserResponse.cs
+++ b/Refresh.GameServer/Endpoints/ApiV3/DataTypes/Response/ApiExtendedGameUserResponse.cs
@@ -26,6 +26,8 @@ public class ApiExtendedGameUserResponse : IApiResponse, IDataConvertableFrom<Ap
     public required bool RpcnAuthenticationAllowed { get; set; }
     public required bool PsnAuthenticationAllowed { get; set; }
     
+    public bool RedirectGriefReportsToPhotos { get; set; } 
+    
     public required string? EmailAddress { get; set; }
     public required bool EmailAddressVerified { get; set; }
     public required bool ShouldResetPassword { get; set; }
@@ -52,6 +54,7 @@ public class ApiExtendedGameUserResponse : IApiResponse, IDataConvertableFrom<Ap
             EmailAddress = user.EmailAddress,
             EmailAddressVerified = user.EmailAddressVerified,
             ShouldResetPassword = user.ShouldResetPassword,
+            RedirectGriefReportsToPhotos = user.RedirectGriefReportsToPhotos,
         };
     }
 
--- a/Refresh.GameServer/Endpoints/Game/ReportingEndpoints.cs
+++ b/Refresh.GameServer/Endpoints/Game/ReportingEndpoints.cs
@@ -3,6 +3,7 @@ using Bunkum.Core.Endpoints;
 using Bunkum.Core.Responses;
 using Bunkum.Listener.Protocol;
 using Bunkum.Protocols.Http;
+using Refresh.GameServer.Authentication;
 using Refresh.GameServer.Database;
 using Refresh.GameServer.Extensions;
 using Refresh.GameServer.Time;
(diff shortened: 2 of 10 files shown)


Refresh.GameServer/Endpoints/ApiV3/DataTypes/Response/ApiExtendedGameUserResponse.cs
```diff
@@ -26,6 +26,8 @@ public class ApiExtendedGameUserResponse : IApiResponse, IDataConvertableFrom<Ap
     public required bool RpcnAuthenticationAllowed { get; set; }
     public required bool PsnAuthenticationAllowed { get; set; }
     
+    public bool RedirectGriefReportsToPhotos { get; set; }
```
jvyden, line 29: Since we declare a new option on the user here, we're gonna need a corresponding `refresh-web` change. Should be trivial enough.

Refresh.GameServer/Endpoints/Game/ReportingEndpoints.cs
```diff
@@ -4,19 +4,69 @@
...
+        {
+            List<SerializedPhotoSubject> subjects = new();
+            if (body.Players != null)
+                subjects.AddRange(body.Players.Select(player => new SerializedPhotoSubject
+                {
+                    Username = player.Username,
+                    DisplayName = player.Username,
+                    //TODO: im not sure what BoundsList expects, seems to be some kind of float value,
+                    //      but im not sure how to get that from a Rect
+                    BoundsList = "",
```
jvyden, line 37: You can pull a couple examples from the tests

Refresh.GameServer/Services/CommandService.cs
```diff
@@ -100,6 +100,11 @@ public void HandleCommand(CommandInvocation command, GameDatabaseContext databas
                 
                 break;
             }
+            case "togglegriefphotos": {
+                user.RedirectGriefReportsToPhotos = !user.RedirectGriefReportsToPhotos;
+                
+                break;
+            }
```
jvyden, line 107: Would prolly work better as an on/off instead of a toggle. That way it's explicit in what you're doing so you can't get since you can't know the result of the command.

It would be a different story if LBP actually used the result of `/filter` on the client, though. I assume they did that so it's harder to know if you hit the moderation filter.

Refresh.GameServer/Endpoints/Game/ReportingEndpoints.cs
Beyley: I do see how its formatted, but im not sure what the values mean, is it `-1 - 1`? `0 - 1`?

jvyden: 0 - 1 i believe

jvyden: As for meaning I'm guessing that it's percents per left/top/bottom/right of the image

Beyley: image
image
After checking my database, it has negative values, so im going to assume its a NDC type deal and its `-1 - 1`

jvyden: I'm guessing 0,0,0,0 is center then?

Beyley: player in bottom left
image
player in top right
image

jvyden (approved):
LGTM, this should be a pretty nice feature :)

jvyden:
Oh, I need to make the refresh-web change first. I'll do that now.

PR #206, opened 2023-10-19
100% test coverage with ActivityEndpoints

Beyley:
This doesnt verify the data *returned* by /lbp/stream is valid but this at least tests that the endpoint doesnt error when it shouldn't.


jvyden:
We can't deserialize, but maybe we can just check that certain keywords exist/don't exist in the raw string?

Beyley:
> We can't deserialize, but maybe we can just check that certain keywords exist/don't exist in the raw string?

hmm that would be quite jank, but would at least give some *basic* protections, i'll do that

jvyden (approved):

Issue #200, opened 2023-10-16
Add a chat command that lets you enter any level


Arcadius2006:
Add a command usable in the ingame chat that can forcibly load a level, regardless of the fact that it's locked or a sub-level

This option would be only available to administrators by default, and level would be accesed by typing in the command proceeded by the level's hash. This could also be modified to accept guids instead, allowing for exploration of levels stored in gamedata

jvyden:
This is especially important now that we have the February 2023 archive. Don't know about guids, but certainly for normal SHA-1 hashes.

jvyden:
Done

PR #210, opened 2023-10-20
Ability for server to override level pages

jvyden:
Will let you easily go to a level from the website (I imagine a button that says 'play now!' ~~that will trigger this) or let you jump into a level by its SHA1/GUID as #200 suggests~~ will come in a separate PR

## TODO
- [x] Define basic API and service for overriding level lists
- [x] Write unit tests
- [x] Expose this system in the website
- [x] Test in-game

### DIFF
--- a/RefreshTests.GameServer/GameServer/TestRefreshGameServer.cs
+++ b/RefreshTests.GameServer/GameServer/TestRefreshGameServer.cs
@@ -1,5 +1,9 @@
+using System.Reflection;
+using Bunkum.Core;
+using Bunkum.Core.Services;
 using Bunkum.Core.Storage;
 using Bunkum.Protocols.Http;
+using JetBrains.Annotations;
 using NotEnoughLogs;
 using NotEnoughLogs.Behaviour;
 using NotEnoughLogs.Sinks;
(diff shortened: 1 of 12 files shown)


jvyden:
okay, okay, time to actually finish this instead of clicking pull changes every couple of days

RefreshTests.GameServer/GameServer/TestRefreshGameServer.cs
```diff
@@ -64,7 +68,17 @@ protected override void SetupServices()
...
+    }
+    
+    [Pure]
+    public TService GetService<TService>() where TService : Service
+    {
+        List<Service> services = (List<Service>)typeof(BunkumServer).GetField("_services", BindingFlags.Instance | BindingFlags.NonPublic)!
+            .GetValue(this._server)!;
+
+        return (TService)services.First(s => typeof(TService) == s.GetType());
     }
```
Beyley, line 83: Im not sure how i feel about this, this should be something exposed in Bunkum, but since its non-critical test code it should be fine, non-blocking for merge but i'd like to see this more cleanly resolved at some point

Beyley (approved):

RefreshTests.GameServer/GameServer/TestRefreshGameServer.cs
jvyden: Yeah, the reasoning behind that behind choice in Bunkum was I wanted to discourage handling of Bunkum's objects, but as Bunkum grew that use-case became more acceptable (think database providers outside of the main BunkumServer for instance). I'll definitely implement this in Bunkum.

PR #222, opened 2023-10-30
API for editing and removing levels

jvyden:


Refresh.GameServer/Endpoints/ApiV3/LevelApiEndpoints.cs
```diff
@@ -66,4 +68,40 @@ public ApiListResponse<ApiLevelCategoryResponse> GetCategories(RequestContext co
...
+        GameLevel? level = database.GetLevelById(id);
+        if (level == null) return ApiNotFoundError.LevelMissingError;
+
+        if (level.Publisher?.UserId != user.UserId) 
+            return ApiAuthenticationError.NoPermissionsForObject;
+
+        database.DeleteLevel(level);
+
+        return new ApiOkResponse();
+    }
```
Beyley, line 106: should this be merged with the admin "delete level" request, with the check on line 100 being ignored if the user is an admin?

jvyden: Probably, but I would prefer to keep admin land and user land entirely separate.

Currently it's a check we handle at the service level rather than the actual endpoint code, so if we checked if the user was an admin for each endpoint then we'd be spamming that check everywhere.

It's just easier to manage in my head.

Beyley (approved):

PR #247, opened 2023-11-03
fixup for dumb bad code not clearing score events when levels are deleted

### LINKED ISSUE #246: Level deletions don't clear score events
image
This is happening because the website is trying to reference a score that doesn't exist, but the event for the score still exists.
image

jvyden:
:(

Closes #246


jvyden:
fuck it

Issue #249, opened 2023-11-03
Ingame Level Filtering


Arcadius2006:
LittleBigPlanet 2, Vita and 3 all let you filter through your search by number of players, tag, game, and text

these do not work
![image](https://github.com/LittleBigRefresh/Refresh/assets/93880348/4113b50a-95ac-43da-ae0c-1657af6ce53f)
please fix!

jvyden:
Why isn't this closed again? Just tags & labels I presume?

Beyley:
> Why isn't this closed again? Just tags & labels I presume?

yup

jvyden:
Closing in favor of #260.

PR #252, opened 2023-11-05
Refactor categories + implement game, player, and self filtering

Beyley:
I'm not 100% happy with the categories refactor, but it was fixes a lot of the type safety and flexibility concerns i was having. There is probably a better way to do this, but i was unable to figure out anything better than "make everything abstract"

### DIFF
--- a/Refresh.GameServer/Database/GameDatabaseContext.Levels.cs
+++ b/Refresh.GameServer/Database/GameDatabaseContext.Levels.cs
@@ -159,32 +159,32 @@ public partial class GameDatabaseContext // Levels
         => this._realm.All<GameLevel>().Where(l => l._Source == (int)GameLevelSource.User).FilterByGameVersion(gameVersion);
 
     [Pure]
-    public DatabaseList<GameLevel> GetLevelsByUser(GameUser user, int count, int skip, TokenGame gameVersion, LevelFilterSettings levelFilterSettings)
+    public DatabaseList<GameLevel> GetLevelsByUser(GameUser user, int count, int skip, LevelFilterSettings levelFilterSettings)
     {
         if (user.Username == DeletedUser.Username)
         {
-            return new DatabaseList<GameLevel>(this.GetLevelsByGameVersion(gameVersion).FilterByLevelFilterSettings(null, levelFilterSettings).Where(l => l.Publisher == null), skip, count);
+            return new DatabaseList<GameLevel>(this.GetLevelsByGameVersion(levelFilterSettings.GameVersion).FilterByLevelFilterSettings(null, levelFilterSettings).Where(l => l.Publisher == null), skip, count);
         }
         
-        return new DatabaseList<GameLevel>(this.GetLevelsByGameVersion(gameVersion).FilterByLevelFilterSettings(null, levelFilterSettings).Where(l => l.Publisher == user), skip, count);
+        return new DatabaseList<GameLevel>(this.GetLevelsByGameVersion(levelFilterSettings.GameVersion).FilterByLevelFilterSettings(null, levelFilterSettings).Where(l => l.Publisher == user), skip, count);
     }
 
     [Pure]
-    public DatabaseList<GameLevel> GetNewestLevels(int count, int skip, TokenGame gameVersion, GameUser? user, LevelFilterSettings levelFilterSettings) =>
-        new(this.GetLevelsByGameVersion(gameVersion)
+    public DatabaseList<GameLevel> GetNewestLevels(int count, int skip, GameUser? user, LevelFilterSettings levelFilterSettings) =>
+        new(this.GetLevelsByGameVersion(levelFilterSettings.GameVersion)
             .FilterByLevelFilterSettings(user, levelFilterSettings)
             .OrderByDescending(l => l.PublishDate), skip, count);
     
     [Pure]
-    public DatabaseList<GameLevel> GetRandomLevels(int count, int skip, TokenGame gameVersion, GameUser? user, LevelFilterSettings levelFilterSettings) =>
-        new(this.GetLevelsByGameVersion(gameVersion)
+    public DatabaseList<GameLevel> GetRandomLevels(int count, int skip, GameUser? user, LevelFilterSettings levelFilterSettings) =>
+        new(this.GetLevelsByGameVersion(levelFilterSettings.GameVersion)
             .FilterByLevelFilterSettings(user, levelFilterSettings)
             .AsEnumerable()
             .OrderBy(_ => Random.Shared.Next()), skip, count);
     
     // TODO: reduce code duplication for getting most of x
     [Pure]
-    public DatabaseList<GameLevel> GetMostHeartedLevels(int count, int skip, TokenGame gameVersion, GameUser? user, LevelFilterSettings levelFilterSettings)
+    public DatabaseList<GameLevel> GetMostHeartedLevels(int count, int skip, GameUser? user, LevelFilterSettings levelFilterSettings)
     {
         IQueryable<FavouriteLevelRelation> favourites = this._realm.All<FavouriteLevelRelation>();
         
@@ -197,13 +197,13 @@ public partial class GameDatabaseContext // Levels
             .Where(l => l != null)
             .Where(l => l._Source == (int)GameLevelSource.User)
             .FilterByLevelFilterSettings(user, levelFilterSettings)
-            .FilterByGameVersion(gameVersion);
+            .FilterByGameVersion(levelFilterSettings.GameVersion);
 
         return new DatabaseList<GameLevel>(mostHeartedLevels, skip, count);
     }
     
     [Pure]
-    public DatabaseList<GameLevel> GetMostUniquelyPlayedLevels(int count, int skip, TokenGame gameVersion, GameUser? user, LevelFilterSettings levelFilterSettings)
+    public DatabaseList<GameLevel> GetMostUniquelyPlayedLevels(int count, int skip, GameUser? user, LevelFilterSettings levelFilterSettings)
     {
         IQueryable<UniquePlayLevelRelation> uniquePlays = this._realm.All<UniquePlayLevelRelation>();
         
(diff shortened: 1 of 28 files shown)


Refresh.GameServer/Endpoints/Game/Levels/FilterSettings/LevelFilterSettings.cs
```diff
@@ -0,0 +1,120 @@
...
+        }
+        
+        string? gameFilterType = context.QueryString.Get("gameFilterType");
+        if (gameFilterType != null)
+            this.GameFilterType = gameFilterType switch
+            {
+                "lbp1" => GameFilterType.LittleBigPlanet1,
+                "lbp2" => GameFilterType.LittleBigPlanet2,
+                "both" => GameFilterType.Both,
+                //NOTE: We dont throw an exception here because an unknown filter setting could very well just be an unhandled case,
```
jvyden, line 89: Ensuring everything stays working is more than acceptable, but I still think we should at least log a warning in this case.

Refresh.GameServer/Endpoints/Game/Levels/FilterSettings/LevelFilterSettings.cs
```diff
@@ -0,0 +1,120 @@
...
+                this.ExcludeMyLevels = false;
+        
+        string? move = context.QueryString.Get("move");
+        if (move != null)
+            this.MoveFilterType = move switch
+            {
+                "true" => MoveFilterType.True,
+                "false" => MoveFilterType.False,
+                "only" => MoveFilterType.Only,
+                //NOTE: We dont throw an exception here because an unknown filter setting could very well just be an unhandled case,
```
jvyden, line 106: Same thing, we should log a warning

Beyley:
This does not fully close #249, as we still dont allow filtering by tags (i was unable to even find the data on the GameLevel object)

Refresh.GameServer/Extensions/LevelEnumerableExtensions.cs
```diff
@@ -28,4 +30,68 @@ public static IQueryable<GameLevel> FilterByGameVersion(this IQueryable<GameLeve
...
+        // MoveFilterType.True => levels,
+        // MoveFilterType.Only => levels.Where(l => l.MoveCompatible),
+        // MoveFilterType.False => levels.Where(l => !l.MoveCompatible),
+        // _ => throw new ArgumentOutOfRangeException()
+        // };
+
+        return levels;
+    }
+    
+    public static IEnumerable<GameLevel> FilterByLevelFilterSettings(this IEnumerable<GameLevel> levels, GameUser? user, LevelFilterSettings levelFilterSettings)
```
jvyden, line 66: We should probably just source generate this, copying the other method. This class is getting stupid. Should be a separate PR tho

Refresh.GameServer/Types/Levels/Categories/DeveloperLevelsCategory.cs
```diff
@@ -0,0 +1,24 @@
...
+using Refresh.GameServer.Database;
+using Refresh.GameServer.Endpoints.Game.Levels.FilterSettings;
+using Refresh.GameServer.Services;
+using Refresh.GameServer.Types.UserData;
+
+namespace Refresh.GameServer.Types.Levels.Categories;
+
+public class DeveloperLevelsCategory : LevelCategory
+{
+    internal DeveloperLevelsCategory() : base("developer", Array.Empty<string>(), false)
```
jvyden, line 12: Uh, what? Shouldn't we be specifying an API route here?

Refresh.GameServer/Types/Levels/Categories/LevelCategory.cs
```diff
@@ -3,86 +3,36 @@
...
-            return (DatabaseList<GameLevel>)this._method.Invoke(database, args.ToArray())!;
-        }
-        catch
-        {
-            Console.WriteLine(this.ApiRoute);
-            throw;
-        }
-    }
+    public abstract DatabaseList<GameLevel>? Fetch(RequestContext context, int skip, int count, MatchService matchService, GameDatabaseContext database, GameUser? user, TokenGame gameVersion,
+        LevelFilterSettings levelFilterSettings);
```
jvyden, line 37: If we're gonna be passing a LevelFilterSettings object everywhere, we might as well move some stuff to it like `gameVersion`. This method's parameter list is very very long. Though, maybe not `skip` and `count` since that would mean it would be pretty annoying to use this API outside of the context of a request

Refresh.GameServer/Endpoints/Game/Levels/FilterSettings/LevelFilterSettings.cs
```diff
@@ -0,0 +1,120 @@
...
+    /// </summary>
+    public string? LabelFilter0 = null;
+    /// <summary>
+    /// The second label filter
+    /// </summary>
+    public string? LabelFilter1 = null;
+    /// <summary>
+    /// The third label filter
+    /// </summary>
+    public string? LabelFilter2 = null;
```
jvyden, line 49: I feel like this should be an array. Any reason this has to be 3 separate fields?

Refresh.GameServer/Endpoints/Game/Levels/FilterSettings/GameFilterType.cs
```diff
@@ -0,0 +1,17 @@
+namespace Refresh.GameServer.Endpoints.Game.Levels.FilterSettings;
+
+public enum GameFilterType
```
jvyden, line 3: Should add xmldoc to explain this is only in LBP2

Refresh.GameServer/Endpoints/Game/Levels/LevelEndpoints.cs
```diff
@@ -38,9 +39,11 @@ public class LevelEndpoints : EndpointGroup
         
         (int skip, int count) = context.GetPageData();
 
+        LevelFilterSettings filterSettings = new(context);
+        
         DatabaseList<GameLevel>? levels = categoryService.Categories
             .FirstOrDefault(c => c.GameRoutes.Any(r => r.StartsWith(route)))?
-            .Fetch(context, skip, count, matchService, database, user, token.TokenGame);
+            .Fetch(context, skip, count, matchService, database, user, token.TokenGame, filterSettings);
```
jvyden, line 46: Should just inline filterSettings like below

Refresh.GameServer/Database/GameDatabaseContext.Levels.cs
```diff
@@ -271,13 +282,14 @@ public DatabaseList<GameLevel> GetBusiestLevels(int count, int skip, TokenGame g
...
+            .FilterByLevelFilterSettings(user, levelFilterSettings)
             .FilterByGameVersion(gameVersion), skip, count);
     }
 
     [Pure]
-    public DatabaseList<GameLevel> SearchForLevels(int count, int skip, TokenGame gameVersion, string query)
+    public DatabaseList<GameLevel> SearchForLevels(int count, int skip, TokenGame gameVersion, LevelFilterSettings levelFilterSettings, string query)
     {
-        IQueryable<GameLevel> validLevels = this.GetLevelsByGameVersion(gameVersion);
+        IQueryable<GameLevel> validLevels = this.GetLevelsByGameVersion(gameVersion).FilterByLevelFilterSettings(null, levelFilterSettings);
```
jvyden, line 292: Why do the other methods include the user, but not this one?

Refresh.GameServer/Database/GameDatabaseContext.Levels.cs
```diff
@@ -244,25 +252,28 @@ public DatabaseList<GameLevel> GetHighestRatedLevels(int count, int skip, TokenG
...
             .Where(l => l.TeamPicked)
+            .FilterByLevelFilterSettings(user, levelFilterSettings)
             .OrderByDescending(l => l.PublishDate), skip, count);
 
     [Pure]
-    public DatabaseList<GameLevel> GetDeveloperLevels(int count, int skip, TokenGame gameVersion) =>
+    public DatabaseList<GameLevel> GetDeveloperLevels(int count, int skip, GameUser? user, LevelFilterSettings levelFilterSettings) =>
         new(this._realm.All<GameLevel>()
             .Where(l => l._Source == (int)GameLevelSource.Story)
+            .FilterByLevelFilterSettings(user, levelFilterSettings)
```
jvyden, line 272: We can probably skip passing the user in here since a developer level will never ever be owned by a user.

jvyden (changes requested):
Yeah, I definitely think it's way better now that we're explicitly calling these functions. This is a big PR so I found it a little hard to keep track of exactly what was going on here, but here are some small comments I found from a relatively quick scroll.

Refresh.GameServer/Types/Levels/Categories/DeveloperLevelsCategory.cs
Beyley: API route is specified, not game route

jvyden: I'm guessing it's handled as a quirk in LevelEndpoints, then?

Refresh.GameServer/Endpoints/Game/Levels/FilterSettings/LevelFilterSettings.cs
Beyley: no reason to separate it, i'll move it to an array

Refresh.GameServer/Types/Levels/Categories/DeveloperLevelsCategory.cs
Beyley: Theres no quirk, the fact the game endpoint array is empty means its just never considered by the LevelEndpoints code

jvyden: I'm surprised it's worked so far, then. I'll open an issue, we can investigate later

Refresh.GameServer/Endpoints/Game/Levels/FilterSettings/LevelFilterSettings.cs
Beyley: What would be the proper way to access the logger from this code? pass it in through the constructor?

jvyden: Good point... I'm not exactly sure. Requiring a logger in the constructor would certainly be sloppy, so I'd say store it statically, but something about that irks me.

It might be okay to just see if this catches on fire. I don't think this will be an issue.

Refresh.GameServer/Extensions/LevelEnumerableExtensions.cs
jvyden: https://github.com/LittleBigRefresh/Refresh/issues/254

jvyden (approved):

PR #265, opened 2023-11-11
Prevent token use across different API types

jvyden:
https://discord.com/channels/1049223665243389953/1049225857350254632/1173004094185668639

### DIFF
--- a/Refresh.GameServer/Authentication/GameAuthenticationProvider.cs
+++ b/Refresh.GameServer/Authentication/GameAuthenticationProvider.cs
@@ -5,6 +5,7 @@ using Refresh.GameServer.Types.UserData;
 using Bunkum.Core.Authentication;
 using Bunkum.Core.Database;
 using Refresh.GameServer.Configuration;
+using Refresh.GameServer.Endpoints;
 using Refresh.GameServer.Types.Roles;
 
 namespace Refresh.GameServer.Authentication;
@@ -30,6 +31,17 @@ public class GameAuthenticationProvider : IAuthenticationProvider<Token>
             tokenType = TokenType.Api;
             tokenData = request.RequestHeaders["Authorization"];
         }
+
+        // ReSharper disable once SwitchExpressionHandlesSomeKnownEnumValuesWithExceptionInDefault
+        string validBaseRoute = tokenType switch
+        {
+            TokenType.Game => GameEndpointAttribute.BaseRoute,
+            TokenType.Api => ApiV3EndpointAttribute.BaseRoute,
+            _ => throw new ArgumentOutOfRangeException(),
+        };
+
+        if (!request.Uri.AbsolutePath.StartsWith(validBaseRoute))
+            return null;
 
         // if still null we dont have a token so bail 
         if (tokenData == null) return null;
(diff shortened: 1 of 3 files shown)


Refresh.GameServer/Authentication/GameAuthenticationProvider.cs
```diff
@@ -31,6 +32,17 @@ public GameAuthenticationProvider(GameServerConfig? config)
             tokenData = request.RequestHeaders["Authorization"];
         }
 
+        // ReSharper disable once SwitchExpressionHandlesSomeKnownEnumValuesWithExceptionInDefault
+        string validBaseRoute = tokenType switch
+        {
+            TokenType.Game => GameEndpointAttribute.BaseRoute,
+            TokenType.Api => ApiV3EndpointAttribute.BaseRoute,
```
Beyley, line 39: Do any api v1 endpoints require auth? if so this will break in that case

RefreshTests.GameServer/Tests/Authentication/TokenAbuseTests.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        using HttpClient gameClient = context.GetAuthenticatedClient(TokenType.Game, user);
+        using HttpClient apiClient = context.GetAuthenticatedClient(TokenType.Api, user);
+        
+        HttpResponseMessage request = apiClient.GetAsync("/lbp/eula").Result;
+        Assert.That(request.StatusCode, Is.EqualTo(Forbidden));
+        
+        request = gameClient.GetAsync("/lbp/eula").Result;
+        Assert.That(request.StatusCode, Is.EqualTo(OK));
+
```
Beyley, line 38: fore matting (WhiteSpace)

Beyley (changes requested):

Refresh.GameServer/Authentication/GameAuthenticationProvider.cs
jvyden: No, our implementation of Lighthouse's API doesn't include any authentication endpoints.

Most of their stuff is done through their Razor front-end which doesn't apply to us.

RefreshTests.GameServer/Tests/Authentication/TokenAbuseTests.cs
jvyden: I don't see what you see

Beyley: extra line of whitespace before the end of the function

jvyden: Oh

Beyley (approved):

Refresh.GameServer/Authentication/GameAuthenticationProvider.cs
jvyden: Should be noted that Lighthouse *does* have authenticated API endpoints, but they're only for sending out invite codes automatically for beta testing. I don't think it would be too difficult to implement in `LegacyApi`, but that one feature is irrelevant to us since we never had that stage so it hasn't been implemented. Another thing is that the token for that feature isn't attached to a user from what I remember correctly so that type of token wouldn't pass through our authentication provider anyways

I wouldn't mind adding authentication support for `LegacyApi` if they ever expand their API, but I don't know if they've thought about doing that in recent times. I think they just don't have a reason to.

PR #275, opened 2023-11-15
Support .NET 8

jvyden:
This change can be made independently of Bunkum, since the .NET 7 version still works under 8.


Beyley (approved):
We should audit the codebase at some point for places where frozen collections would make sense

jvyden:
Yeah, I'm doing that right now - I have a couple commits I haven't pushed that I'll push to master instead.

PR #285, opened 2023-11-20
Abstract room-keeping

### LINKED ISSUE #282: RFC: Move matching to an optional microservice
# The problem
As Refresh adapts to be scalable and redundant, it needs to start adapting from the stateful mess it's currently in. One of these stateful area that will cause significant issues is the matching system. Rooms are stored in the main `Refresh.GameServer` project's memory, and is not synced or shared anywhere.

When we spin up multiple instances of the server, match requests will get load balanced randomly and the room states will eventually desynchronize between pods. This is okay for a single-server instance, but not for larger instances like LBR.

# Implementation
To solve this, we can add a new project to the solution named `Refresh.MatchServer`. It will be a Bunkum server with a similar structure to the current APIv3 used in `Refresh.GameServer`. There will be no authentication, as this server will be expected to be kept inaccessible.

## REST API:
The match server will expose a handful of REST endpoints to provide basic management. Logic will stay inside the application. This can be thought of as a data store.

### Managing rooms
`GET /api/v1/rooms`: Get all rooms on the server.
`GET /api/v1/rooms/{uuid}`: Get a room by an ID.
`PUT /api/v1/rooms/{uuid}`: Add a room to the server by an ID.
`DELETE /api/v1/rooms/{uuid}`: Delete a room from the server by an ID.
`PATCH /api/v1/rooms/{uuid}`: Update a room on the server by an ID.

### Finding rooms by a user
`GET /api/v1/users/uuid/{uuid}`: Get a room by a user's ID.
`GET /api/v1/users/username/{username}`: Get a room by a user's name.

## Keeping our one-click philosophy
Obviously, moving everything to a separate server would mean that we would require extra configuration. We want to avoid this as much as possible, so when we implement this it should behind an interface. The interface is defined below.

```csharp
public interface IRoomAccessor
{
    public ApiList<GameRoom> GetRooms(int skip, int count);
    public GameRoom? GetRoomByUuid(ObjectId uuid);
    public void AddRoom(GameRoom room);
    public void RemoveRoom(GameRoom room);
    public void UpdateRoom(GameRoom room);

    public GameRoom? GetRoomByUser(GameUser user) => this.GetRoomByUserUuid(user.UserId);
    public GameRoom? GetRoomByUserUuid(ObjectId uuid);
    public GameRoom? GetRoomByUsername(string username);
}
```

Then, we add a new property to our configuration to `refreshGameServer.json`:

```jsonc
{
    "externalMatchServer": null, // Null would mean use a in-memory-based IRoomAccessor
    "externalMatchServer": "http://matchServer/", // Setting a URL would mean to use a network-based IRoomAccessor.
}
```

Beyley:
This PR will __not__ satisfy #282, this is a partial implementation to step towards that, an implementation of `Refresh.MatchServer` can come in a later PR. 
(im bikeshedding here, but a better name would be `Refresh.RoomServer` or `Refresh.RoomTracker`, since it doesnt actually do matching, only tracking of rooms)

A couple questions:
 - Should `AddRoom` and `UpdateRoom` be combined into a single function? They practically do the same thing. (and `UpdateRoom` is implemented by calling `AddRoom`, since `AddRoom` makes sure the UUID isnt already being used)
 - Should operations like `match.RoomAccessor.GetAllRooms().Count()` be exposed through the REST API? I feel like sending the whole room list over the network is wasteful when we could sum it on the room storage server and send the sum back
 - How are force matches going to work? should we move them to the database?

A couple notes: 
 - The new implementation is fully thread safe (the old `MatchService` was not). 
 - `InMemoryRoomAccessor` attempts to replicate the behavior that a transactional REST API would provide, so it makes copies of the list.
 - Multiple functions that were on `MatchService` have been superseded by functions on `RoomAccessor`, and external callers of said functions have been updated to call into the underlying `RoomAccessor` directly.
 - Tests were updated to match the new transactional API, since they made some dirty assumptions before.


jvyden:
> Should `AddRoom` and `UpdateRoom` be combined into a single function?

I suppose we can. I was thinking of `PUT` vs `PATCH` when I wrote that. But I suppose we could combine everything the two a `POST` request and then return either 201 Created or 200 OK.

> Should operations like `match.RoomAccessor.GetAllRooms().Count()` be exposed through the REST API? 

Absolutely yes, I hadn't even considered that. Whoops. Are there any other common queries we might want to expose?

> How are force matches going to work? should we move them to the database?

Good question. Force matching should be set on the GameUser.

jvyden:
I'll add an addendum to the proposal to handle common queries

jvyden (approved):
I'll approve for now with the old proposal in mind but it should definitely be updated to reflect the new addendum.

jvyden:
Bump, how's progress on this?

Beyley:
> Bump, how's progress on this?

have a partial implementation of the appendix in my local changes, just havent finished it yet

Refresh.GameServer/Types/Matching/RoomStatistics.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace Refresh.GameServer.Types.Matching;
+
+[JsonObject(NamingStrategyType = typeof(CamelCaseNamingStrategy))]
+public class RoomStatistics
+{
+    public ushort PlayerCount { get; set; }
+    public ushort PlayersInPodCount { get; set; }
+    public ushort RoomCount { get; set; }
+    public Dictionary<TokenGame, ushort> PerGame { get; set; }
```
jvyden, line 11: These are causing null warnings that shouldn't happen ever, let's just suppress them at the top with `#nullable disable`

jvyden (commented):
Just some nitpicks. Been a while since I've thought about this so it's not exactly a full conscious review but I'm sure we'll cross that bridge when we get to it.

jvyden (approved):

PR #288, opened 2023-11-25
Switch to MailKit for SMTP

### LINKED ISSUE #287: Migrate to MailKit for SMTP
`SmtpClient`, the method our implementation is currently using has apparently been [deprecated](https://github.com/dotnet/platform-compat/blob/master/docs/DE0005.md) and feature-locked for the last 5 years. In .NET 8, it seems to have broken.

Microsoft's recommended alternative is [MailKit](https://github.com/jstedfast/MailKit).

jvyden:
Closes #287

### DIFF
--- a/Refresh.GameServer/Services/SmtpService.cs
+++ b/Refresh.GameServer/Services/SmtpService.cs
@@ -1,7 +1,8 @@
-using System.Net;
-using System.Net.Mail;
 using Bunkum.Core;
 using Bunkum.Core.Services;
+using MailKit.Net.Smtp;
+using MailKit.Security;
+using MimeKit;
 using NotEnoughLogs;
 using Refresh.GameServer.Configuration;
 using Refresh.GameServer.Types.UserData;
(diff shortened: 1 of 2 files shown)


Refresh.GameServer/Services/SmtpService.cs
```diff
@@ -20,33 +20,36 @@ public class SmtpService : EndpointService
...
+        if (!this._integrationConfig.SmtpEnabled) return false;
+
+        using SmtpClient client = new();
+
+        SecureSocketOptions tlsOptions = SecureSocketOptions.None;
+
+        if (this._integrationConfig is { SmtpTlsEnabled: true, SmtpPort: 587 })
+            tlsOptions = SecureSocketOptions.StartTls;
+        else if (this._integrationConfig.SmtpTlsEnabled)
+            tlsOptions = SecureSocketOptions.SslOnConnect;
```
Beyley, line 36: Is there a reason that we hardcode only 587 for TLS?
This behaviour was not in the old code

jvyden: 587 is plaintext only and requires the `STARTTLS` extension, but other ports are TLS \*always\*.

The old client probably had similar logic internally.

Beyley (approved):

PR #306, opened 2023-12-24
ResourceHelper: Add MIP importing

Beyley:
This adds full importing of LBP PSP MIP files.

The LBP PSP asset decryption key must be placed in the `keys` folder next to the executable (create it if missing), with the name `psp`. If the key is missing, Refresh isnt even able to detect whether or not a file is a MIP file.

This also brings in a new dependency for LZO decompression, since MIP files are sometimes compressed.

@jvyden https://github.com/LittleBigRefresh/Refresh/commit/9a91cb7b1087c4f19b8462641064cb9a6926054b this commit im unsure about, should it be the client's or the API's responsibility to append the `psp/` for PSP assets?


jvyden:
Regarding 9a91cb7b1087c4f19b8462641064cb9a6926054b, probably the API's responsibility. It should be as easy as possible to write a client.

Refresh.GameServer/Importing/Mip/MipMetadata.cs
```diff
@@ -0,0 +1,6 @@
+namespace Refresh.GameServer.Importing.Mip;
+
+public class MipMetadata
+{
+    
+}
```
jvyden, line 6: ?

Refresh.GameServer/Refresh.GameServer.csproj
```diff
@@ -19,6 +20,11 @@
     
     <PropertyGroup Condition=" '$(Configuration)' == 'Release' ">
         <PublishSingleFile>true</PublishSingleFile>
+        <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    </PropertyGroup>
+    
+    <PropertyGroup Condition=" '$(Configuration)' == 'Debug' ">
+      <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
jvyden, line 27: All this should be at the leading `PropertyGroup` at the top of the file.

Refresh.GameServer/Refresh.GameServer.csproj
```diff
@@ -65,6 +71,7 @@
       <PackageReference Include="BCrypt.Net-Next" Version="4.0.3" />
       <PackageReference Include="CommandLineParser" Version="2.9.1" />
       <PackageReference Include="Discord.Net.Webhook" Version="3.13.0" />
+      <PackageReference Include="IronCompress" Version="1.5.1" />
```
jvyden, line 74: This library mentions a lack of support for Linux arm32 for LZO, which I believe is one of our targets? Kinda? I'm unsure if this is actually used in production anywhere but it was pushed for by @Slendy when adding Docker support so it's something to consider

Refresh.GameServer/Resources/ResourceHelper.cs
```diff
@@ -8,5 +14,124 @@ public static Stream StreamFromResource(string name)
...
+        //Decompress the data
+        using IronCompressResult decompressed = Iron.Value!.Decompress(Codec.LZO, buf, (int)len);
+
+        Xor(data, key);
+
+        //Return a copy of the decompressed data
+        return decompressed.AsSpan().ToArray();
+    }
+    
+    public static byte[] XorAlloc(ReadOnlySpan<byte> data, ReadOnlySpan<byte> key)
```
jvyden, line 116: Is this function used anywhere?

Refresh.GameServer/Resources/ResourceHelper.cs
```diff
@@ -8,5 +14,124 @@ public static Stream StreamFromResource(string name)
...
+        }
+    }
+
+    private static void XorScalar(Span<byte> data, ReadOnlySpan<byte> key)
+    {
+        //Iterate over all bytes
+        for (int i = 0; i < data.Length; i++) 
+            //Iterate over all data, and XOR it with the key
+            data[i] ^= key[i];
+    }
```
jvyden, line 68: Unsure how it affects performance here but this could probably be inlined

jvyden (changes requested):
How are we ripping the XOR key from the game? We should definitely document that step on the wiki.

Additionally, what's the plan for production? Are we planning to just run a re-import of assets? I'm not sure the importer is ready for that - IIRC it only looks at files not present in the database.

Refresh.GameServer/Resources/ResourceHelper.cs
Beyley: In the case of the scalar algorithm its going to be neglidgible if not unnocticable performance wise compared to the hot loop, this function isnt called often enough for inlining to make a meaningful performance improvement.

Refresh.GameServer/Resources/ResourceHelper.cs
Beyley: Not atm, its a leftover, its unneeded now, i'll remove it

Refresh.GameServer/Importing/Mip/MipMetadata.cs
Beyley: This is an ImageSharp thing, it wants a metadata class, but since we have nothing to fill it out with, its left empty, you can see the reference [here](https://github.com/LittleBigRefresh/Refresh/pull/306/files#diff-8e2e55d3c11ef30dcab4a28cc7ed1bfa8ee1d5374e7e96e0e2ef3bcf729472d3)

Refresh.GameServer/Refresh.GameServer.csproj
Beyley: rider moment, will fix

Refresh.GameServer/Refresh.GameServer.csproj
Beyley: This is a valid concern, we should open an issue upstream about arm32 linux (and open a tracking issue in Refresh aswell), but given this is a non-essential feature which is disabled by default, i dont feel trying to find an ARM32 solution should be blocking for merging. If required, the server owner could install the required native library themselves too (in this case being [`minilzo`](https://github.com/yuhaoth/minilzo/tree/b088e80336ef11c5a29ec949a950d4f0fec9ec05))

Beyley:
> How are we ripping the XOR key from the game? We should definitely document that step on the wiki.

I'll have to write a guide for that, i dont have one on-hand.

> Additionally, what's the plan for production? Are we planning to just run a re-import of assets? I'm not sure the importer is ready for that - IIRC it only looks at files not present in the database.

From my testing, a `-i` then a `-I` is sufficient to trigger the conversion to take place.

Beyley:
Marking as draft, since complete rewrite of decryption is needing to be done, see [this conversation](https://discord.com/channels/1049223665243389953/1049225857350254632/1189776523339771974)

Beyley:
Going to mark as ready for review, since the Refresh side code is now done. We still need to write the tool for extracting the key (should probably go into docs repo so its on the docs wiki?), but i think that should be done separately. For now, if you want to extract the key, manually follow the instructions [here](https://discord.com/channels/1049223665243389953/1049225857350254632/1189776523339771974) using a LBP PSP US 2.0.5 copy of the game's EBOOT

jvyden (approved):

Refresh.GameServer/Refresh.GameServer.csproj
jvyden: Actually, I think arm32 isn't worth chasing - if it becomes a problem then we'll handle it. As it stands, everyone ever is dropping 32-bit support apart from Raspbian who still prominently features arm32 over arm64. But I think that's about the only case.

PR #317, opened 2024-01-02
GameUser: Split Vita icon hash from PS3 icon hash

Beyley:


Refresh.GameServer/Database/GameDatabaseProvider.cs
```diff
@@ -168,6 +168,9 @@ protected override void Migrate(Migration migration, ulong oldVersion)
 
             // Version was bumped here to delete invalid favourite level relations
             if (oldVersion < 101) migration.NewRealm.RemoveRange(newUser.FavouriteLevelRelations.Where(r => r.Level == null));
+
+            // In version 102 we split the Vita icon hash from the PS3 icon hash
+            if (oldVersion < 102) newUser.VitaIconHash = oldUser.IconHash;
```
jvyden, line 173: Since these systems are incompatible we should reset to `0` here; it's not too big of a deal if we just ask people to reset their avatars on Vita.

Refresh.GameServer/Database/GameDatabaseContext.Users.cs
```diff
@@ -100,12 +100,38 @@ public void UpdateUserData(GameUser user, SerializedUpdateData data, TokenGame g
...
+                        //If the icon is a remote asset, then it will work on Vita as well, so set the Vita hash 
+                        if (!data.IconHash.StartsWith('g'))
+                        {
+                            user.VitaIconHash = data.IconHash;
+                        }
+#endif
+                        user.IconHash = data.IconHash;
+                        break;
+                    case TokenGame.LittleBigPlanetVita:
+#if false //Enable this code once https://github.com/LittleBigRefresh/Refresh/issues/309 is resolved
```
jvyden, line 119:
```suggestion
#if false // TODO: Enable this code once https://github.com/LittleBigRefresh/Refresh/issues/309 is resolved
                        //If the icon is a remote asset, then it will work on Vita as well, so set the Vita hash 
                        if (!data.IconHash.StartsWith('g'))
                        {
                            user.VitaIconHash = data.IconHash;
                        }
#endif
                        user.IconHash = data.IconHash;
                        break;
                    case TokenGame.LittleBigPlanetVita:
#if false // TODO: Enable this code once https://github.com/LittleBigRefresh/Refresh/issues/309 is resolved
```

jvyden (changes requested):

PR #327, opened 2024-01-03
Refactor RefreshGameServer to support multiple server types

jvyden:
# Why
I am currently starting the beginnings of the Room Server discussed in #282. While creating the new project I noticed I was rewriting a lot of code regarding management and bootstrapping, so I decided to do a bit of refactoring.

# How
I created a new project named `Refresh.Common` which contains a new `RefreshServer` class. The normal `RefreshGameServer` extends from this.

`RefreshServer` bootstraps Bunkum and logging functionality, and `RefreshGameServer` now solely manages the workers and injecting of services into Bunkum.

# Remarks
Had to do some weird stuff regarding initialization due to some Bunkum weirdness that came back to bite me in the ass, but it all works out in the end.

### DIFF
--- a/Refresh.GameServer/RefreshGameServer.cs
+++ b/Refresh.GameServer/RefreshGameServer.cs
@@ -10,9 +10,7 @@ using Bunkum.Core.Storage;
 using Bunkum.HealthChecks;
 using Bunkum.HealthChecks.RealmDatabase;
 using Bunkum.Protocols.Http;
-using NotEnoughLogs;
-using NotEnoughLogs.Behaviour;
-using NotEnoughLogs.Sinks;
+using Refresh.Common;
 using Refresh.GameServer.Authentication;
 using Refresh.GameServer.Configuration;
 using Refresh.GameServer.Database;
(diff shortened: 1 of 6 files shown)


Refresh.GameServer/RefreshGameServer.cs
```diff
@@ -255,7 +224,7 @@ public void SetAdminFromEmailAddress(string emailAddress)
         context.SetUserRole(user, GameUserRole.Admin);
     }
 
-    public void Dispose()
+    public override void Dispose()
     {
         this.Logger.Dispose();
         this._databaseProvider.Dispose();
```
Beyley, line 230: Some of this code is duplicated from the base implementation, you can just do `base.Dispose()` instead of duplicating the `Logger.Dispose` and `GC.SuppressFinalize`

Beyley (changes requested):

Refresh.GameServer/RefreshGameServer.cs
jvyden: True, didn't realize I had duplicated this

Beyley (approved):

Issue #309, opened 2023-12-28
Some games do not understand other image formats


Beyley:
- PS3 misreads PSP's TGA/MIP
![image](https://github.com/LittleBigRefresh/Refresh/assets/34501060/07d04a1e-5498-4f52-b49f-0de8b50306b8)
- Vita misreads PS3's TEX
- PSP misreads PNG/TEX files

PR #328, opened 2024-01-03
Start cropping+scaling images for the right contexts

### LINKED ISSUE #270: Crop level/user icon on website and webhook to 1:1
Since the final product will always be square, it is useless to keep the full image data.

crop them automatically where applicable

Beyley:
Fixes #270 and #309


Refresh.GameServer/Types/Assets/GameAsset.cs
```diff
@@ -1,6 +1,15 @@
...
 using Realms;
+using Refresh.GameServer.Authentication;
 using Refresh.GameServer.Database;
+using Refresh.GameServer.Importing;
+using Refresh.GameServer.Importing.Mip;
+using Refresh.GameServer.Resources;
 using Refresh.GameServer.Types.UserData;
+using SixLabors.ImageSharp;
+using SixLabors.ImageSharp.PixelFormats;
+using SixLabors.ImageSharp.Processing;
```
jvyden, line 12: Putting all this image processing stuff directly in GameAsset feels disgusting. Maybe we can move this to a partial instead?

Refresh.GameServer/Types/Assets/GameAsset.cs
```diff
@@ -23,4 +32,341 @@ public partial class GameAsset : IRealmObject
...
+
+    /// <summary>
+    /// Transforms an input image into a photo-friendly format
+    /// </summary>
+    /// <param name="image">The input photo</param>
+    /// <returns>The transformed image</returns>
+    private Image<Rgba32>? PhotoTransformation(Image image)
+    {
+        //No transformation needs to be done for photos.
+        return null;
```
jvyden, line 50: This function definitely shouldn't be here. If we ever have the need to do transformations on photos then we should just write the function then.

Refresh.GameServer/Types/Assets/GameAsset.cs
```diff
@@ -23,4 +32,341 @@ public partial class GameAsset : IRealmObject
...
+                        //Return the existing asset hash
+                        return this.AssetHash;
+                    }
+
+                    //Set the image to use to the source image
+                    image = sourceImage;
+                }
+
+                //Save the image as a PNG file in a byte array in memory
+                MemoryStream memory = new();
```
jvyden, line 137: That variable name makes me think `Memory<T>`, not `MemoryStream`. In refresh we usually use `ms` or just `stream`.

Refresh.GameServer/Types/Assets/GameAsset.cs
```diff
@@ -23,4 +32,341 @@ public partial class GameAsset : IRealmObject
...
+                //Return the new icon hash
+                return convertedHash;
+            }
+            case TokenGame.LittleBigPlanetPSP: {
+                Image<Rgba32> sourceImage = decodeImage(dataStorePath);
+
+                //Transform the image, if no transformation is needed, use the source image
+                Image<Rgba32> image = transformImage(sourceImage) ?? sourceImage;
+
+                MemoryStream memory = new();
```
jvyden, line 156: Same here regarding `memory`

Refresh.GameServer/Types/Assets/GameAsset.cs
```diff
@@ -23,4 +32,341 @@ public partial class GameAsset : IRealmObject
...
+        }
+    }
+
+    public string GetAsPhoto(TokenGame game, GameDatabaseContext database, IDataStore dataStore)
+    {
+        return this.GetAsGeneric(
+            game,
+            database,
+            dataStore,
+            this.PhotoTransformation,
```
jvyden, line 181: See my comment above

```suggestion
            () => null,
```

Refresh.GameServer/Types/Assets/GameAsset.cs
```diff
@@ -23,4 +32,341 @@ public partial class GameAsset : IRealmObject
...
+            dataStore,
+            this.CropToIcon,
+            () => this.AsMainlineIconHash,
+            hash => database.SetAsMainlineIconHash(this, hash),
+            () => this.AsMipIconHash,
+            hash => database.SetAsMipIconHash(this, hash)
+        );
+    }
+
+    public string GetAsGeneric(TokenGame game, GameDatabaseContext database, IDataStore dataStore, Func<Image, Image<Rgba32>?> transformImage, Func<string?> getMainline, Action<string> setMainline, Func<string?> getMip, Action<string> setMip)
```
jvyden, line 203: Some xmldoc on this would be nice at least. No idea how to review this function.

Refresh.GameServer/Types/Assets/GameAsset.cs
```diff
@@ -23,4 +32,341 @@ public partial class GameAsset : IRealmObject
...
+            case GameAssetType.Material:
+            case GameAssetType.Mesh:
+            case GameAssetType.Palette:
+            case GameAssetType.Script:
+            case GameAssetType.MoveRecording:
+            case GameAssetType.VoiceRecording:
+            case GameAssetType.SyncedProfile:
+            case GameAssetType.Unknown:
+            default:
+                throw new ArgumentOutOfRangeException();
```
jvyden, line 369:
```suggestion
                throw new InvalidOperationException($"Format '{this.AssetType}' is not a valid image.");
```

RefreshTests.GameServer/RefreshTests.GameServer.csproj
```diff
@@ -49,6 +49,11 @@
 
     </ItemGroup>
 
+    <ItemGroup>
+      <None Remove="Resources\1x1.png" />
+      <EmbeddedResource Include="Resources\1x1.png" />
```
jvyden, line 54: Why two?

Refresh.GameServer/Database/GameDatabaseContext.Assets.cs
```diff
@@ -23,4 +42,22 @@ public partial class GameDatabaseContext // AssetConfiguration
...
+            asset.AsMainlineIconHash = hash;
+        });
+    
+    public void SetAsMipIconHash(GameAsset asset, string hash) =>
+        this._realm.Write(() =>
+        {
+            asset.AsMipIconHash = hash;
+        });
+    
+    public void SetAsMainlinePhotoHash(GameAsset asset, string hash) =>
```
jvyden, line 58: Name makes lots more sense if you drop the 'As', e.g. `SetMainlinePhotoHash`

jvyden (changes requested):
Well, that was a pain in the ass to review, but I did it. Still haven't done any local testing myself - this is the first pass.

Will come back with a second pass after I test locally in a bit.

Refresh.GameServer/Importing/Mip/MipEncoder.cs
```diff
@@ -0,0 +1,121 @@
...
+    protected override void Encode<TPixel>(Image<TPixel> origImage, Stream stream, CancellationToken cancellationToken)
+    {
+        Image<Rgba32> image = origImage.CloneAs<Rgba32>();
+        //Quantize the image into 256 colors
+        image.Mutate(ctx => ctx.Quantize(new WuQuantizer(new QuantizerOptions
+        {
+            Dither = ErrorDither.Atkinson,
+            MaxColors = 256,
+        })));
+        image.SaveAsPng("rawquant.png");
```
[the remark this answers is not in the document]
Beyley: correct

Refresh.GameServer/Importing/Importer.cs
```diff
@@ -15,7 +15,8 @@ public abstract class Importer
 {
     private readonly Logger _logger;
     protected readonly Stopwatch Stopwatch;
-    protected readonly Lazy<byte[]?> PSPKey;
+    
+    public static Lazy<byte[]?> PSPKey;
```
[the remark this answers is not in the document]
jvyden: Oops, thought I deleted this one.

RefreshTests.GameServer/RefreshTests.GameServer.csproj
Beyley: idk rider does that

jvyden: Gotcha. I think only the second is necessary

jvyden:
Strange error while to migrate on my dev instance:

```
[01/04/24 00:38:43] [Debug] [Startup] Initializing database provider...
[01/04/24 00:38:43] [Debug] [Startup] Starting listener...
[01/04/24 00:38:43] [Info] [Startup] Listening...
[01/04/24 00:38:43] [Debug] [Startup] Warming up database provider...
Unhandled exception. System.AggregateException: Exception occurred in a Realm.MigrationCallback callback. (The best overloaded method match for 'string.this[int]' has some invalid arguments)
 ---> Microsoft.CSharp.RuntimeBinder.RuntimeBinderException: The best overloaded method match for 'string.this[int]' has some invalid arguments
   at CallSite.Target(Closure, CallSite, Object, Range)
   at System.Dynamic.UpdateDelegates.UpdateAndExecute2[T0,T1,TRet](CallSite site, T0 arg0, T1 arg1)
   at Refresh.GameServer.Database.GameDatabaseProvider.Migrate(Migration migration, UInt64 oldVersion) in /home/jvyden/Documents/Refresh/Refresh.GameServer/Database/GameDatabaseProvider.cs:line 316
   at Realms.SharedRealmHandle.OnMigration(IntPtr oldRealmPtr, IntPtr newRealmPtr, IntPtr migrationSchema, Schema oldSchema, UInt64 schemaVersion, IntPtr managedConfigHandle)
   --- End of inner exception stack trace ---
   at Realms.NativeException.ThrowIfNecessary()
   at Realms.SharedRealmHandle.Open(Configuration configuration)
   at Realms.RealmConfiguration.CreateHandle(Configuration& configuration)
   at Realms.RealmConfigurationBase.CreateRealm()
   at Realms.Realm.GetInstance(RealmConfigurationBase config)
   at Bunkum.RealmDatabase.RealmDatabaseProvider`1.GetContext()
   at Refresh.GameServer.Database.GameDatabaseProvider.Warmup() in /home/jvyden/Documents/Refresh/Refresh.GameServer/Database/GameDatabaseProvider.cs:line 79
   at Bunkum.Core.BunkumServer.RunStartupTasks()
   at Bunkum.Core.BunkumServer.Start(Nullable`1 taskOverride)
   at Refresh.GameServer.RefreshGameServer.Start() in /home/jvyden/Documents/Refresh/Refresh.GameServer/RefreshGameServer.cs:line 156
   at Program.<Main>$(String[] args) in /home/jvyden/Documents/Refresh/Refresh.GameServer/Startup.cs:line 22
   at Program.<Main>(String[] args)

Process finished with exit code 134.
```

Have you tested with the fresh instance of a dev package?

Beyley:
> Have you tested with the fresh instance of a dev package?

I thought i did but i guess not. this is just C# being C#, pushed a fix

Refresh.GameServer/Importing/Mip/MipEncoder.cs
```diff
@@ -0,0 +1,121 @@
...
+            int y = 0;
+            int xStart = 0;
+            int xTarget = blockWidth;
+            int yStart = 0;
+            int yTarget = blockHeight;
+            
+            int bytesToWrite = image.Width * image.Height;
+            for (int i = 0; i < bytesToWrite; i++)
+            {
+                #region hack to get swizzled coordinates
```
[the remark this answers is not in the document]
Beyley: > Same thing, probably a new function for this. `UnswizzleCoordinates()`?

Given the code is inherently stateful, im not sure how to turn this into an isolated function, unless i want to pass in a bunch of locals as ref parameters

Refresh.GameServer/Importing/Mip/MipEncoder.cs
```diff
@@ -0,0 +1,121 @@
...
+        
+        //Fill in the header's CLUT
+        foreach ((Rgba32 key, byte value) in outputPalette) header.Clut[value] = key;
+
+        //Write the header
+        header.Write(stream);
+        
+        BinaryWriter writer = new(stream);
+
+        {
```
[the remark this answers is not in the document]
Beyley: Scope is there because C# is silly and says that the `x` and `y` shadow the `x` and `y` used above while getting the palette (even though its in its own scope), i'll move the scope to a function

Refresh.GameServer/Resources/ResourceHelper.cs
```diff
@@ -9,19 +10,57 @@ namespace Refresh.GameServer.Resources;
...
-        throw new NotImplementedException(); //TODO
+        byte[] initialCounter = new byte[16];
+        key[..8].CopyTo(initialCounter);
+        
+        //If the file is big enough, try to compress it
+        bool compress = data.Length > 1024;
+
+        //If we want to compress the data, do so, if not, dont
+        ReadOnlySpan<byte> resultData = compress
+            ? Iron.Value!.Compress(Codec.LZO, data, null, CompressionLevel.SmallestSize).AsSpan().ToArray()
```
jvyden, line 38: So... we have a byte array from `Compress`. Then, we cast it to a span. Then we convert the span to a byte array. Finally, we store it as a ReadOnlySpan<byte>.

What?

Refresh.GameServer/Importing/Mip/MipEncoder.cs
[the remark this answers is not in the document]
jvyden: If it's too much trouble then I guess it's fine as-is.

Refresh.GameServer/Resources/ResourceHelper.cs
Beyley: Compress does not return a byte array, its a special data type, calling `AsSpan()` gets the span form of the underlying data, then calling `ToArray` makes a copy, since we only need to read from that copy, we then store that into a `ReadOnlySpan<byte>`
In this case though we can skip the copy step, since we dont actually return `resultData` instead copying it into `final`, i'll remove that extra `ToArray`

jvyden:
Getting lots of errors like this locally while trying to browse levels via the API:

```csharp
[01/04/24 19:04:59] [Error] [Request] System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
 ---> System.InvalidOperationException: Format 'Unknown' is not a valid image.
   at Refresh.GameServer.Types.Assets.GameAsset.GetAsGeneric(TokenGame game, GameDatabaseContext database, IDataStore dataStore, Func`2 transformImage, Func`1 getMainline, Action`1 setMainline, Func`1 getMip, Action`1 setMip) in /home/jvyden/Documents/Refresh/Refresh.GameServer/Types/Assets/GameAsset.Conversion.cs:line 360
   at Refresh.GameServer.Types.Assets.GameAsset.GetAsIcon(TokenGame game, GameDatabaseContext database, IDataStore dataStore) in /home/jvyden/Documents/Refresh/Refresh.GameServer/Types/Assets/GameAsset.Conversion.cs:line 165
```

Maybe we should try to handle this more softly?

Refresh.GameServer/Resources/ResourceHelper.cs
jvyden: Should also make consts for the byte values down below (e.g. 0x19, 0x10, etc.) below just to give them some names.

Beyley:
`Unknown` is quite odd to be hitting, is it not? Especially if its often, what assets in your database are marked as "unknown" and are also used as icons?

But yeah more gracefully handling it is smart (returning `"0"` maybe?)

jvyden:
What's likely happening is I'm missing the PSP key and it's trying to convert PSP levels' icons.

I could see this being hit often.

jvyden:
> (returning `"0"` maybe?)
I'd say returning the original asset hash is better? Worst case scenario is the asset is garbled if the destination platform cant read the format, but best case scenario the platform can and it comes out as it would in master.

Beyley:
a! yeah thats probably it, lets handle this gracefully, and log something in console similar to our network settings xml file, where it only logs it *once* the first time its hit.
maybe with the message `Unable to convert texture with type "Unknown", you may be missing the PSP decryption key!`

And yeah returning the original asset hash would prevent some issues (and cant really be abused, since "Unknown" can only come from PSP, and non-psp cant access PSP assets anyway since they are in a different folder)

jvyden:
Missing the PSP key is something we can know beforehand, so I'd prefer if problems relating to it just be warned about once on startup. (also this is unrelated, but I doubt the warnings for the endpoints you mentioned are actually genuinely helpful and should probably be removed)

If we're trying to read an Unknown asset then we're fucked anyways so we should just silently fail.

jvyden (approved):

Beyley:
> If we're trying to read an Unknown asset then we're fucked anyways so we should just silently fail.

Missing PSP key should already be logged, so i think all we need to do is just remove the exception and return the existing asset hash if conversion fails

Beyley:
All review comments addressed

jvyden:
I'll test web locally again just to see and then merge.

jvyden (approved):
Good

PR #333, opened 2024-01-05
Filter out sub-levels from level results.

### LINKED ISSUE #332: Sub-levels show up in level browser
Sub-levels, which are supposed to be unviewable by players other than the level creator, still show up in the in-game search results

they do not appear on the earth itself, but they CAN be played instead of the main level, so it should be patched immedietaly

Beyley:
Fixes #332


jvyden (changes requested):
Hmm, authors should *definitely* be able to view their own sublevels. Otherwise they wouldn't be able to find & delete/modify them.

PR #335, opened 2024-01-06
Show converted icons in category level previews and activity

Beyley:
`.Select` is an _interesting_ footgun i did not consider until now

### DIFF
--- a/Refresh.GameServer/Endpoints/ApiV3/DataTypes/Response/Admin/ApiAdminQueuedRegistrationResponse.cs
+++ b/Refresh.GameServer/Endpoints/ApiV3/DataTypes/Response/Admin/ApiAdminQueuedRegistrationResponse.cs
@@ -23,5 +23,5 @@ public class ApiAdminQueuedRegistrationResponse : IApiResponse, IDataConvertable
         };
     }
 
-    public static IEnumerable<ApiAdminQueuedRegistrationResponse> FromOldList(IEnumerable<QueuedRegistration> oldList) => oldList.Select(FromOld)!;
+    public static IEnumerable<ApiAdminQueuedRegistrationResponse> FromOldList(IEnumerable<QueuedRegistration> oldList) => oldList.Select(FromOld).ToList()!;
 }
--- a/Refresh.GameServer/Endpoints/ApiV3/DataTypes/Response/ApiActivityPageResponse.cs
+++ b/Refresh.GameServer/Endpoints/ApiV3/DataTypes/Response/ApiActivityPageResponse.cs
@@ -42,5 +42,5 @@ public class ApiActivityPageResponse : IApiResponse, IDataConvertableFrom<ApiAct
         return response;
     }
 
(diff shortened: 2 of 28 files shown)


jvyden (changes requested):
what the fuck am i looking at? this is a mishmash of like 3 different changes all of which are completely and utterly unexplained

Beyley:
> what the fuck am i looking at? this is a mishmash of like 3 different changes all of which are completely and utterly unexplained

`.ToList` changes are to create copies of the list, as currently the direct result of these functions are used, and when attempting to iterate that and change it, [copies are made and the actual valuse arent updated](https://discord.com/channels/1049223665243389953/1049225857350254632/1193301066213773373), the change from returning AssetHash to returning null is to give outer bits of code more control over the value (eg. in the API we were dropping `psp/` in places because of this).
The change from `"0"` to `this.AssetHash` in places is to fallback to a more reasonable value upon failure (which also coincided with the last change)

jvyden (approved):
After a good hour of trying to understand this, cross-referencing with our current codebase I still don't get what this PR is doing exactly or what was broken in the first place, so I'm just going to merge seeing as it does appear to do *something* somehow.

I really do recommend writing unit tests for universal assets, since I struggle to follow the flow with it.

Beyley:
> After a good hour of trying to understand this, cross-referencing with our current codebase I still don't get what this PR is doing exactly or what was broken in the first place, so I'm just going to merge seeing as it does appear to do _something_ somehow.

In the case of category level previews being broken, we just werent calling the `WithExtraData` method, in the case of the activity levels being broken, its a much more subtle bug.

When you get an enumerator from a `.Select` call, every time you enumerate that, you get a new set of values, eg
```
var enumerator = something.Select(SomeTransformation);

//SomeTransformation is called for every item in `something` here, creating new objects
foreach(var val in enumerator) {
	...
}

//SomeTransformation is called for every item in `something` here, creating new objects again
foreach(var val in enumerator) {
	...
}
```

This meant that when running [this](https://github.com/LittleBigRefresh/Refresh/blob/687ccd0ea7e24103c381020769aa0af589339684/Refresh.GameServer/Endpoints/ApiV3/DataTypes/Response/ApiActivityPageResponse.cs#L32) line of code, `this.Levels` gets enumerated, and new `ApiGameLevelResponse` objects are being created, with `FillInExtraData` is being called on those new objects.

When the serializer later enumerates `this.Levels`, a new set of objects are created using the transformation in `.Select`, meaning the changes done before in `FillInExtraData` are never applied from the serializers perspective.

By calling `.ToList()`, we are doing that enumeration of `.Select` once, and storing those values into a `List<T>`, now when enumerating that `List<T>`, you are always enumerating the same objects, which means both `FillInExtraData` and the serializer both are using the same objects.

should have probably included this in the original PR, but was lazy. This technically extended to many more places than just the universal assets stuff, its just that the dropped `FillInExtraData` calls werent visually obvious with the other cases.

PR #341, opened 2024-01-13
Implement recent activity filters, and show less types to users in-game

Beyley:
Only the first half is done so far, we still show the whole server recent activity log to all users. The only current new changes is the implementation of the filters.

### DIFF
--- a/Refresh.GameServer/Database/GameDatabaseContext.Activity.cs
+++ b/Refresh.GameServer/Database/GameDatabaseContext.Activity.cs
@@ -28,6 +28,24 @@ public partial class GameDatabaseContext // Activity
             .Where(e => e.Timestamp < timestamp && e.Timestamp >= endTimestamp)
             .AsEnumerable();
 
+        if (user != null)
+        {
+            List<ObjectId?> favouriteUsers = user.UsersFavourited.AsEnumerable().Select(f => (ObjectId?)f.UserToFavourite.UserId).ToList();
+            List<ObjectId?>? userFriends = friendService?.GetUsersFriends(user, this)?.Select(u => (ObjectId?)u.UserId).ToList();
+
+            query = query.Where(e =>
+                e.User.UserId == user.UserId ||
+                e.StoredObjectId == user.UserId ||
+                favouriteUsers.Contains(e.User.UserId) ||
+                favouriteUsers.Contains(e.StoredObjectId) ||
+                (userFriends?.Contains(e.User.UserId) ?? false) ||
+                (userFriends?.Contains(e.StoredObjectId) ?? false) ||
+                this.GetLevelById(e.StoredSequentialId ?? int.MaxValue)?.Publisher?.UserId == user.UserId ||
+                e.EventType == EventType.Level_TeamPick ||
+                e.EventType == EventType.User_FirstLogin
+            );
+        }
+        
         if (excludeMyLevels && user != null)
         {
             //Filter the query to events which either arent level related, or which the level publisher doesnt contain the user
@@ -53,7 +71,7 @@ public partial class GameDatabaseContext // Activity
         {
             query = query.Where(e => e.User.UserId != user.UserId && e.StoredObjectId != user.UserId);  
         }
-        
+
         return new(query.OrderByDescending(e => e.Timestamp), skip, count);
     }
 
@@ -74,7 +92,8 @@ public partial class GameDatabaseContext // Activity
         IEnumerable<Event> query = this._realm.All<Event>()
             .Where(e => e._StoredDataType == 1 && e.StoredSequentialId == level.LevelId)
             .Where(e => e.Timestamp < timestamp && e.Timestamp >= endTimestamp)
-            .AsEnumerable();
+            .AsEnumerable()
+            .OrderByDescending(e => e.Timestamp);
         
         if (excludeFriends && user != null && friendService != null)
         {
@@ -96,8 +115,7 @@ public partial class GameDatabaseContext // Activity
             query = query.Where(e => e.User.UserId != user.UserId && e.StoredObjectId != user.UserId);  
         }
         
-        return new DatabaseList<Event>(query
-            .OrderByDescending(e => e.Timestamp), skip, count);
+        return new DatabaseList<Event>(query, skip, count);
     }
 
     public int GetTotalEventCount() => this._realm.All<Event>().Count();
--- a/Refresh.GameServer/Database/GameDatabaseContext.Activity.cs
+++ b/Refresh.GameServer/Database/GameDatabaseContext.Activity.cs
@@ -12,110 +12,95 @@ public partial class GameDatabaseContext // Activity
 {
     [Pure]
     public DatabaseList<Event> GetRecentActivity(
-        int count, 
-        int skip, 
-        long timestamp, 
-        long endTimestamp,
-        bool excludeMyLevels = false, 
-        bool excludeFriends = false,
-        bool excludeFavouriteUsers = false,
-        bool excludeMyself = false,
-        GameUser? user = null,
+        ActivityQueryParameters parameters,
         FriendStorageService? friendService = null
     )
     {
         IEnumerable<Event> query = this._realm.All<Event>()
-            .Where(e => e.Timestamp < timestamp && e.Timestamp >= endTimestamp)
+            .Where(e => e.Timestamp < parameters.Timestamp && e.Timestamp >= parameters.EndTimestamp)
             .AsEnumerable();
 
-        if (user != null)
+        if (parameters.User != null)
         {
-            List<ObjectId?> favouriteUsers = user.UsersFavourited.AsEnumerable().Select(f => (ObjectId?)f.UserToFavourite.UserId).ToList();
-            List<ObjectId?>? userFriends = friendService?.GetUsersFriends(user, this)?.Select(u => (ObjectId?)u.UserId).ToList();
+            List<ObjectId?> favouriteUsers = parameters.User.UsersFavourited.AsEnumerable().Select(f => (ObjectId?)f.UserToFavourite.UserId).ToList();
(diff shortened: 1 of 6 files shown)


Refresh.GameServer/Database/GameDatabaseContext.Activity.cs
```diff
@@ -28,6 +28,24 @@ public partial class GameDatabaseContext // Activity
...
+            query = query.Where(e =>
+                e.User.UserId == user.UserId ||
+                e.StoredObjectId == user.UserId ||
+                favouriteUsers.Contains(e.User.UserId) ||
+                favouriteUsers.Contains(e.StoredObjectId) ||
+                (userFriends?.Contains(e.User.UserId) ?? false) ||
+                (userFriends?.Contains(e.StoredObjectId) ?? false) ||
+                this.GetLevelById(e.StoredSequentialId ?? int.MaxValue)?.Publisher?.UserId == user.UserId ||
+                e.EventType == EventType.Level_TeamPick ||
+                e.EventType == EventType.User_FirstLogin
```
jvyden, line 45: Some of this should *definitely* be moved to the main `IQueryable` call in at the top of the function if possible. This is going to be a performance nightmare.

Refresh.GameServer/Endpoints/Game/ActivityEndpoints.cs
```diff
@@ -54,7 +54,71 @@ public class ActivityEndpoints : EndpointGroup
...
+        bool excludeMyself = bool.Parse(context.QueryString["excludeMyself"] ?? "false");
+        
+        string? tsStr = context.QueryString["timestamp"];
+        string? tsEndStr = context.QueryString["endTimestamp"];
+        if (tsStr != null && !long.TryParse(tsStr, out timestamp)) return BadRequest;
+        if (tsEndStr != null && !long.TryParse(tsEndStr, out endTimestamp)) return BadRequest;
+
+        if (endTimestamp == 0) endTimestamp = timestamp - 86400000 * 7; // 1 week
+
+        ActivityPage page = ActivityPage.LevelActivity(database, 20, 0, timestamp, endTimestamp, level, excludeFriends, excludeFavouriteUsers, excludeMyself, user, friendStorageService);
```
jvyden, line 80: Feel like it's better to pass in a `ref struct` or something similar for all these arguments. Maybe call it `ActivityQueryParameters` or something?

Refresh.GameServer/Types/Activity/ActivityPage.cs
```diff
@@ -53,6 +53,78 @@ public ActivityPage()
...
+        bool excludeFriends = false,
+        bool excludeFavouriteUsers = false,
+        bool excludeMyself = false,
+        GameUser? user = null,
+        FriendStorageService? friendStorageService = null
+    )
+    {
+        if (timestamp == 0) timestamp = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
+
+        DatabaseList<Event> events = database.GetRecentActivityForLevel(level, count, skip, timestamp, endTimestamp, excludeFriends, excludeFavouriteUsers, excludeMyself, user, friendStorageService);
```
jvyden, line 72: Possible null reference warning

Refresh.GameServer/Types/Activity/ActivityPage.cs
```diff
@@ -53,6 +53,78 @@ public ActivityPage()
...
+
+        ActivityPage page = new()
+        {
+            Events = new List<Event>(events.Items),
```
jvyden, line 76: Why do we clone here?

jvyden (changes requested):
Code quality & performance concerns about this one.

Refresh.GameServer/Database/GameDatabaseContext.Activity.cs
Beyley: > Some of this should _definitely_ be moved to the main `IQueryable` call in at the top of the function if possible. This is going to be a performance nightmare.

Cant be on IQueryable, indirect accesses (like `e.User.UserId`) are unsupported by Realm

Refresh.GameServer/Endpoints/Game/ActivityEndpoints.cs
Beyley: Yeah i was thinking that aswell, the parameters ballooned a lot, will do this

Refresh.GameServer/Database/GameDatabaseContext.Activity.cs
jvyden: Some of these queries, yes, but not all. `e.StoredObjectId`, `e.EventType` checks for example can be moved earlier.

Refresh.GameServer/Types/Activity/ActivityPage.cs
Beyley: copy/paste error, this code is copied from the constructor, which does copy, but copies for a different reason, will remove the copy

Refresh.GameServer/Database/GameDatabaseContext.Activity.cs
jvyden: If its a big issue, then I recommend extracting the filtering logic to a new function, and then split this function in to 3 separate ones for levels, users, and global queries. I think that just makes more sense in general, too.

Beyley: Im not sure how to write this logic in a fashion where i *can* split it up into multiple `.Where` calls. This filters to where the event type is *any* of these many possible things. I'd have to do two `.Where` calls (both on the raw query, cant be chained), then do `.Union` to create a union of the two queries, which im not sure how much more performant that would be, given now im enumerating three times (one per `.Where`, one for the `.Union`).

jvyden: Right, makes sense. I'm vehemently against how long that takes (250ms is frankly absurd especially given the power of your workstation) but if there's no better way to do it then I guess it'll have to work.

This is something that we'll have to consider when we get PostgreSQL support in a functional state. This makes me wonder if we can do raw Realm queries somehow.

That said, we can still the refactoring I suggested. Bunching together logic for what's ultimately 3 different types of result is just a disaster.

Beyley: Yeah performance im really not happy with (in release mode even after the JIT warms up its ~100ms). Just to be sure i understand what you mean with the refactor, do you mean having separate functions for "get filtered global recent activity", "get filtered recent activity scoped to this user", "get filtered global recent activity for a level"?

jvyden: Yep.

Refresh.GameServer/Database/GameDatabaseContext.Activity.cs
```diff
@@ -1,23 +1,103 @@
...
+        FriendStorageService? friendService = null)
+    {
+        if (parameters.Timestamp == 0) 
+            parameters.Timestamp = DateTimeOffset.UtcNow.ToUnixTimeMilliseconds();
+        
+        IEnumerable<Event> query = this._realm.All<Event>()
+            .Where(e => e.Timestamp < parameters.Timestamp && e.Timestamp >= parameters.EndTimestamp)
+            .AsEnumerable();
+
+        if (parameters is { ExcludeMyLevels: true, User: not null })
```
jvyden, line 52: This style of syntax is a little untrustworthy. Is this function backed by unit tests?

Refresh.GameServer/Database/GameDatabaseContext.Activity.cs
```diff
@@ -1,23 +1,103 @@
...
+                e.EventType == EventType.Level_TeamPick ||
+                e.EventType == EventType.User_FirstLogin
+            );
+        }
+        
+        return new(query.OrderByDescending(e => e.Timestamp), parameters.Skip, parameters.Count);
+    }
+
+    private IEnumerable<Event> GetRecentActivity(ActivityQueryParameters parameters,
+        FriendStorageService? friendService = null)
```
jvyden, line 43: Why is this optional? Shouldn't this be required?

Refresh.GameServer/Database/GameDatabaseContext.Activity.cs
```diff
@@ -1,23 +1,103 @@
...
     
     [Pure]
-    public DatabaseList<Event> GetRecentActivityForLevel(GameLevel level, int count, int skip, long timestamp, long endTimestamp) => 
-        new(this._realm.All<Event>()
+    public DatabaseList<Event> GetGlobalRecentActivity(
+        ActivityQueryParameters parameters,
+        FriendStorageService? friendService = null
+    )
+    {
+        return new(this.GetRecentActivity(parameters, friendService).OrderByDescending(e => e.Timestamp), parameters.Skip, parameters.Count);
```
jvyden, line 87:
```suggestion
        return new DatabaseList<Event>(this.GetRecentActivity(parameters, friendService).OrderByDescending(e => e.Timestamp), parameters.Skip, parameters.Count);
```

jvyden (changes requested):
Part 1 of my review. My brain isn't fully here with us today so I'll take a second proper look tomorrow. :pensive:

Refresh.GameServer/Database/GameDatabaseContext.Activity.cs
Beyley: we dont want FriendStorageServices passed in when viewing someone elses recent activity (which you can do ingame), since then you could use that to kinda RE someone's PSN friends

Refresh.GameServer/Database/GameDatabaseContext.Activity.cs
Beyley: Its not backed by unit tests to my knowledge, but this logic was generated using Rider's refactoring tools from the original logic, and it produces better IL, which is why i used it.

jvyden (approved):

PR #345, opened 2024-01-18
Prevent using /play to see invalid levels on wrong games.

Beyley:


Arcadius2006:
imo, this should only apply to regular users

admins having the ability to do that actually seems kind of fitting, if not downright useful

Beyley:
> imo, this should only apply to regular users
> 
> admins having the ability to do that actually seems kind of fitting, if not downright useful

This usecase is already covered by the force token game command.

Arcadius2006:
oh okay

Refresh.GameServer/Services/LevelListOverrideService.cs
```diff
@@ -37,14 +38,61 @@ public void AddOverridesForUser(GameUser user, IEnumerable<GameLevel> levels)
...
+                    break;
+                //Allow all for website
+                case TokenGame.Website:
+                    break;
+                default:
+                    throw new ArgumentOutOfRangeException();
+            }
+            
+            levels.Add(level);
+        }
```
jvyden, line 94: This definitely shouldn't belong here. Pretty sure we have this logic duplicated multiple times across the codebase - maybe we can move it somewhere static? Something like `bool IsVersionCompatible(GameVersion currentGame, GameVersion levelGame)`

jvyden (changes requested):

Refresh.GameServer/Services/LevelListOverrideService.cs
Beyley: I swore this was the case aswell, but [this](https://github.com/LittleBigRefresh/Refresh/blob/e52849d7841c23ee60a8eab8b07feccd48e73962/Refresh.GameServer/Extensions/LevelEnumerableExtensions.cs#L11) is the closest duplicate code i can find (but even then a static function wouldnt help for LevelEnumerableExtensions, since IQueryable shenanigans)

But extracting this code into a static function i will do anyway for the future incase we need this logic again (quite likely)

jvyden (approved):

PR #346, opened 2024-01-19
Add option to unescape XML for filter requests.

Beyley:
This allows you to use special tags like `<em>`, `<font>`, `<player>` for advanced text editing.

This is already possible by modifying plan files or by using a proxy to intercept the /filter requests, this option just makes it *much* more accessible to users.

Requires a change in refresh-web.


jvyden (changes requested):
Code looks fine, but this should be a command in-game similar to Grief Report Photos. The toggle in `refresh-web` probably won't be implemented since I'm focusing efforts on CosmosFE

jvyden (approved):

PR #348, opened 2024-01-20
Remove `FriendStorageService` and replace the system with mutuals

### LINKED ISSUE #239: Automatically heart PSN friends

### LINKED ISSUE #344: Remove `FriendStorageService` in favor of mutuals
Related: #239

jvyden:
Closes #344, closes #239

# Why
This is another bit of state inside Refresh that we need to get rid of as friend data is stored in-memory. It also came up as a potential annoyance during the implementation of recent activity filters (#341)

# How
Since friend data only counts for users registered to Refresh, we can use the existing `FavouriteUserRelation` system to determine friends instead. That way, friend status is determined entirely by the state of the database.

Since a mutual now essentially means a friend, I've taken the liberty of implementing #239, making `POST /npdata` just heart the list of friends the game gives us.

# Remarks
The database operation to retrieve a list of mutuals is unfortunately quite slow due to Realm's limited query ability. This shouldn't affect things too heavily, since it should be a semi-rare query - only really called during Recent Activity and the occasional time someone visits the "my friends" category in LBP.

This can technically be optimized significantly by adding a new relation type that marks a mutual relationship, but I'd rather do this after we figure out Postgres since we'd probably be able to just do it in a single query without introducing new logic given how EF's query builder (and ability to do raw queries on the database) is much more advanced than Realm.

Lastly, this **entirely** removes `FriendStorageService` in-case that wasn't obvious.


Beyley (approved):
lgtm

PR #351, opened 2024-01-20
Resolve most build warnings

jvyden:
# Remarks
Realm's source generator still generates some nullability warnings (17 as of this PR), but unfortunately we don't have any reasonable control over them. This PR just resolves the warnings that we *can* resolve.


Beyley (approved):
I wish we could easily enable "warnings as errors" but as per [the docs](https://learn.microsoft.com/en-us/visualstudio/code-quality/how-to-suppress-code-analysis-warnings-for-generated-code?view=vs-2022) it is not possible to ignore warnings from generated code, so sadly we'll just have to keep an eye out in the future to keep warnings low.

PR #354, opened 2024-01-25
Send locked status in minimal response

Beyley:
Before this patch, you could play a locked level by clicking on it from the level browse list before it loads the full information.


jvyden (changes requested):
This should include other, similar flags (sublevel status and shareable in particular)

jvyden (approved):

PR #355, opened 2024-01-26
Revert "Filter out sub-levels from level results"

jvyden:
#333 broke the ability for level authors to view their own sublevels.

In my opinion, it's better to show them wrongly than to hide them wrongly, so let's just revert this for now.


Beyley:
agreed, will fix this properly at some point in the future, we should re-open https://github.com/LittleBigRefresh/Refresh/issues/332 aswell.

Beyley (approved):

PR #368, opened 2024-02-26
Add the ability for beta builds to have their own isolated space

jvyden:
Some beta builds can end up posting levels that can crash players playing on mainline versions of the game. For example, a player playing on a deploy build can upload a level that will be registered as an LBP1 level. This causes a hard crash upon loading in LBP1.

This also introduces 2 commands that are __not__ debug commands:
- `/beta`: Switches the token to be `TokenGame.BetaBuild`. Can be used for builds that do not properly identify themselves as beta builds, e.g. deploy.
- `/revoketoken`: Revokes the token, essentially forcing the game to redo the handshake with the server. This can be used to switch away from beta mode, can also just be generally useful.

This was vaguely tested.

### DIFF
--- a/Refresh.GameServer/Endpoints/Game/DataTypes/Response/GameUserResponse.cs
+++ b/Refresh.GameServer/Endpoints/Game/DataTypes/Response/GameUserResponse.cs
@@ -117,6 +117,7 @@ public class GameUserResponse : IDataConvertableFrom<GameUserResponse, GameUser>
             TokenGame.LittleBigPlanetVita => old.VitaPlanetsHash,
             TokenGame.LittleBigPlanetPSP => "0",
             TokenGame.Website => "0",
+            TokenGame.BetaBuild => old.Lbp2PlanetsHash, // TODO: Planet hash for beta builds
             _ => throw new ArgumentOutOfRangeException(nameof(gameVersion), gameVersion, null),
         };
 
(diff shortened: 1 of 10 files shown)


Refresh.GameServer/Extensions/LevelEnumerableExtensions.cs
```diff
@@ -36,14 +38,17 @@ public static IQueryable<GameLevel> FilterByLevelFilterSettings(this IQueryable<
...
+        {
+            levels = levelFilterSettings.GameFilterType switch {
+                GameFilterType.LittleBigPlanet1 => levels.Where(l => l._GameVersion == (int)TokenGame.LittleBigPlanet1),
+                GameFilterType.LittleBigPlanet2 => levels.Where(l => l._GameVersion == (int)TokenGame.LittleBigPlanet2),
+                //NOTE: ideally this should be .Where(l => l._GameVersion == (int)TokenGame.LittleBigPlane1 || l._GameVersion == (int)TokenGame.LittleBigPlane2)
+                //      however, there should be no differences in all real-world cases
+                GameFilterType.Both => levels,
+                _ => throw new ArgumentOutOfRangeException(),
+            };
+        }
```
Beyley, line 51: It should be commented on why this filtering is disabled for beta builds, i was confused when i first read this code

Refresh.GameServer/Endpoints/Game/DataTypes/Response/GameUserResponse.cs
```diff
@@ -130,6 +131,7 @@ private void FillInExtraData(GameUser old, TokenGame gameVersion, GameDatabaseCo
                 //Fill out LBP2/LBP1 levels
                 goto case TokenGame.LittleBigPlanet2;
             }
+            case TokenGame.BetaBuild:
```
Beyley, line 134: Given the beta builds are fully isolated, shouldnt this logic be different than LBP2?

Beyley (changes requested):
only some small things

Beyley (approved):
code looks good, needs a rebase though

PR #382, opened 2024-03-27
Allow server owners to specify contact information

jvyden:
Servers owner can now specify some essential contact information. Primarily we want name and email, and optionally Discord information.

Name and email **requires** support in refresh-web, but Discord can be excluded for now.


Beyley (approved):
lgtm

PR #386, opened 2024-03-29
Ability to rename a user from the CLI

jvyden:


Beyley (approved):
this information is good

PR #391, opened 2024-03-31
Make all our slot endpoints slotType agnostic

Beyley:
This de-duplicates a lot of code, and also implements a bunch of currently unimplemented endpoints.

This changes behavior somewhat, but the changes are not important (returning `NotFound` for invalid slot types, rather than `BadRequest`). I can see arguments for both of these being valid to respond with, but its simplest to return NotFound here.

### DIFF
--- a/Refresh.GameServer/Database/GameDatabaseContext.Levels.cs
+++ b/Refresh.GameServer/Database/GameDatabaseContext.Levels.cs
@@ -351,24 +351,18 @@ public partial class GameDatabaseContext // Levels
     [Pure]
     public GameLevel? GetLevelByIdAndType(string slotType, int id)
     {
-        GameLevel? level;
-        
         switch (slotType)
         {
             case "user":
-                level = this.GetLevelById(id);
-                break;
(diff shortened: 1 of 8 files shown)


Refresh.GameServer/Database/GameDatabaseContext.Levels.cs
```diff
@@ -348,6 +348,29 @@ public DatabaseList<GameLevel> SearchForLevels(int count, int skip, GameUser? us
...
+                if (id < 0)
+                    return null;
+                
+                level = this.GetStoryLevelById(id);
+                break;
+            default:
+                return null;
+        }
+        
+        return level;
```
jvyden, line 371: Don't see why you couldn't just return directly:
```suggestion
        switch (slotType)
        {
            case "user":
                return this.GetLevelById(id);
            case "developer":
                if (id < 0)
                    return null;
                
                return this.GetStoryLevelById(id);
            default:
                return null;
        }
```

jvyden (changes requested):

PR #404, opened 2024-04-15
Fix a couple flaws with Cool Levels

jvyden:
I've noticed that users sometimes see a new level and downvote it, immediately causing the level's CR to go negative. I fixed this by giving a bonus score of 20 points to allow new levels more chance to survive in Cool Levels.

Decay is still a thing of course, so these points won't last forever.

I also noticed a restricted user who still has levels uploaded being at the top of Cool Levels, so I'm hoping to put a stop to that with a penalty. We should reconsider showing these levels in *all* level lists. Maybe not for restricted users, but especially for banned users - but that's not within scope for this PR.


Beyley (approved):
epiku mute

PR #405, opened 2024-04-15
Add new "SafeMedia" tier between Safe and PotentiallyUnwanted

jvyden:
This PR introduces a new tier for asset safety levels named "SafeMedia". It's intended to represent assets that are images or audio to help administrators single out images during content blocking.

Hopefully, this should replace our need to disable asset uploads on production.

### DIFF
--- a/Refresh.GameServer/Types/Assets/AssetSafetyLevel.cs
+++ b/Refresh.GameServer/Types/Assets/AssetSafetyLevel.cs
@@ -5,15 +5,20 @@ public enum AssetSafetyLevel
     /// <summary>
     /// This asset is used in normal gameplay/operation and is okay to be uploaded.
     /// </summary>
+    /// <seealso cref="SafeMedia"/>
     Safe = 0,
+    /// <summary>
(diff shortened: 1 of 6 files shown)


Refresh.GameServer/Types/Assets/AssetSafetyLevel.cs
```diff
@@ -5,15 +5,20 @@ public enum AssetSafetyLevel
     /// <summary>
     /// This asset is used in normal gameplay/operation and is okay to be uploaded.
     /// </summary>
+    /// <seealso cref="SafeMedia"/>
     Safe = 0,
     /// <summary>
+    /// This asset is still used in normal gameplay similar to <see cref="Safe"/>, and is okay to be uploaded, but the distinction can be useful.
```
Beyley, line 11: this comment is hard to follow and doesnt actually explain what it means by media

Beyley (changes requested):

Refresh.GameServer/Types/Assets/AssetSafetyLevel.cs
jvyden: I got ChatGPT to do it because I'm tired

jvyden:
Not approved but don't care

PR #407, opened 2024-04-16
Several improvements to logging

jvyden:
- Shows role instead of user id in logs, e.g. `jvyden420 (Admin)`
- Logs authentication successes, e.g. when a user logs in, logs out, or refreshes a token
- Logs assets that were uploaded and who uploaded them


Beyley (approved):
ni li pona mute tawa mi. sina o pana e sitelen ni tawa jan ale!

PR #419, opened 2024-04-21
Implement ability to mark level as a reupload

### LINKED ISSUE #233: Ability to mark a level as a reupload

turecross321:
This is not finished yet, but I am still submitting this PR because I have discovered a problem that I am not entirely sure how we should tackle.

Fixes #233 

What has been implemented:
- API V3 admin route to mark a level as a reupload, and the ability to set the `OriginalPublisher`
- Add `IsReUpload` and `OriginalPublisher` to the v3 API level response

As for the Game API, the plan was for the server to, instead of returning the reuploader's username as the publisher when viewing their reupload, return `OriginalPublisher`'s username, like this:

image

This works well when viewing the level from the reuploader's earth, but problems arise when viewing the level from elsewhere. This is because the game tries to fetch the `OriginalPublisher` user, and their other levels -- which doesn't go very well since the entire point of this is that `OriginalPublisher` does not have an account on the server.

From my testing in LBP2, it leads to an excessively long loading screen until it eventually gives up and shows the level, but I wouldn't be surprised if it leads to worse behavior in other games such as LBP Vita, which typically crashes when it doesn't receive what it expects.

image

I see two solutions to this problem; we can either simply return the publisher as `!DeletedUser` and put the original publisher's username in the level description, or we make a system that, instead of returning 404's when the game requests `OriginalPublisher`'s profile/levels, returns what makes the game happy enough to not throw a fit.

The first solution would be significantly easier and less prone to something going wrong, but it would certainly be cooler with the other solution.

What are our thoughts?

### DIFF
--- a/Refresh.GameServer/Database/GameDatabaseContext.Levels.cs
+++ b/Refresh.GameServer/Database/GameDatabaseContext.Levels.cs
@@ -180,7 +180,7 @@ public partial class GameDatabaseContext // Levels
 
         if (user.Username.StartsWith("!"))
         {
-            string withoutPrefix = user.Username.TrimStart('!');
+            string withoutPrefix = user.Username[1..];
             return new DatabaseList<GameLevel>(this.GetLevelsByGameVersion(levelFilterSettings.GameVersion).FilterByLevelFilterSettings(null, levelFilterSettings).Where(l => l.OriginalPublisher == withoutPrefix), skip, count);
         }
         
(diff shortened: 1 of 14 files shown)


jvyden:
My personal idea for how this would have worked is having fake users generated on the fly, e.g. `!${OriginalPublisher}`

Such users aren't actually persisted in the database much like how `!DeletedUser` works so they don't conflict with PSN/RPCN usernames and don't take up useless space.

jvyden (commented):
Code looks good from an early morning glance.

turecross321:
Something like this? This seems to work well 👍👍

image

jvyden:
Should be. :+1:

turecross321:
Cool, I think this should be ready for review then.

jvyden:
Tests seem to be failing?

turecross321:
My bad -- NOW it should be ready for review

jvyden (approved):

jvyden:
Normally I'd wait for beyley to review, but she needs a fkn break after that shit

PR #420, opened 2024-04-21
Send score when overtaken

### LINKED ISSUE #232: Notifications for score overtakes

turecross321:
Fixes #232

I haven't really done enough testing with this, because I lack the amount of people to do it, but I THINK all is well

The FontAwesome icon `medal` doesn't seem to work on the website for some reason? But that seems like a client issue and not a server issue

image
image
image

### DIFF
--- a/RefreshTests.GameServer/Tests/Levels/ScoreLeaderboardTests.cs
+++ b/RefreshTests.GameServer/Tests/Levels/ScoreLeaderboardTests.cs
@@ -402,7 +402,7 @@ public class ScoreLeaderboardTests : GameServerTest
         using TestContext context = this.GetServer();
         const int userAmount = 10;
 
-        List<GameUser> users = Array.Empty<GameUser>().ToList();
+        List<GameUser> users = new(userAmount);
         for (int i = 0; i < userAmount; i++)
         {
             users.Add(context.CreateUser());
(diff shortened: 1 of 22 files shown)


jvyden:
Seems okay to me. You should write unit tests for this if you want to be sure, though.

jvyden:
`refresh-web` requires icons to be imported manually, by the way. It won't show up until you add it to the icon library. The rewrite fixes this.

RefreshTests.GameServer/Tests/Levels/ScoreLeaderboardTests.cs
```diff
@@ -394,4 +395,60 @@ public void DoesntPlayLevelWithCountOnMainline()
...
         Assert.That(message.StatusCode, Is.EqualTo(BadRequest));
     }
+
+    [Test]
+    public void OvertakeNotificationsWorkProperly()
+    {
+        using TestContext context = this.GetServer();
+        const int userAmount = 10;
+
+        List<GameUser> users = Array.Empty<GameUser>().ToList();
```
jvyden, line 405:
```suggestion
        List<GameUser> users = new List<GameUser>(usersAmount);
```

RefreshTests.GameServer/Tests/Levels/ScoreLeaderboardTests.cs
```diff
@@ -394,4 +395,60 @@ public void DoesntPlayLevelWithCountOnMainline()
...
+                if (notification.Title != "Score overtaken") continue;
+                
+                notificationSent = true;
+                context.Database.DeleteNotification(notification);
+            }
+            
+            if (i - 1 == 0) // Notification should not be sent if the last #1 entry had a score of 0
+                Assert.That(notificationSent, Is.EqualTo(false));
+            else
+                Assert.That(notificationSent, Is.EqualTo(true));
```
jvyden, line 435: Having assertions behind flags like this is a little sketch, I won't lie. Can you add checks to make sure that these are both run?

RefreshTests.GameServer/Tests/Levels/ScoreLeaderboardTests.cs
```diff
@@ -394,4 +395,60 @@ public void DoesntPlayLevelWithCountOnMainline()
...
+
+            // Check that notification was not sent to people who weren't previously #1
+            if (i <= 1) continue;
+            
+            notificationRecipient = users[i - 2];
+
+            notificationSent = false;
+            foreach (GameNotification notification in notificationRecipient.Notifications)
+            {
+                if (notification.Title != "Score overtaken") continue;
```
jvyden, line 445: This condition could break if the title changes. Are there other notifications being sent conflicting with your test?

jvyden (changes requested):

RefreshTests.GameServer/Tests/Levels/ScoreLeaderboardTests.cs
turecross321: Not right now, but I figured that it prevents the test from breaking if other notifications get implemented later

Maybe this could be improved if we make a `NotificationType` enum?

RefreshTests.GameServer/Tests/Levels/ScoreLeaderboardTests.cs
turecross321: Would replacing that with the following suffice?
`Assert.That(notificationSent, Is.EqualTo(i - 1 != 0));`

RefreshTests.GameServer/Tests/Levels/ScoreLeaderboardTests.cs
turecross321: Should I assume that the overtake notification is the only one that's been sent for now then?

jvyden: Yeah.

RefreshTests.GameServer/Tests/Levels/ScoreLeaderboardTests.cs
turecross321: Right, gotcha

turecross321:
Should be good now.

jvyden (approved):

PR #421, opened 2024-04-22
API Support for Contests

### LINKED ISSUE #416: Contests
- Custom banner per-contest on the website's front page, akin to the Steam Store's "takeover"s
    image
- New category per contest that filters by a contests' 'tag' (e.g. `#LBPCC1`)
- Show running contests in `/announce`
- Ability to view past contests
- Ability for users to be selected to manage their contest (but only admins may create them)

jvyden:
Implements and closes #416

Someone is planning to run a contest with us soon, let's get this supported for them :)

### DIFF
--- a/Refresh.GameServer/Types/Levels/Categories/ContestCategory.cs
+++ b/Refresh.GameServer/Types/Levels/Categories/ContestCategory.cs
@@ -15,15 +15,21 @@ public class ContestCategory : LevelCategory
         this.Description = "Levels from a contest.";
         this.FontAwesomeIcon = "certificate";
         this.IconHash = "g820608";
-        this.Hidden = true;
     }
     
     public override DatabaseList<GameLevel>? Fetch(RequestContext context, int skip, int count, MatchService matchService,
         GameDatabaseContext database, GameUser? user, LevelFilterSettings levelFilterSettings)
     {
+        // try to find a contest by the query parameter
         string? contestId = context.QueryString["contest"];
         GameContest? contest = database.GetContestById(contestId);
-        if (contest == null) return null;
+        
+        // if we cant find one by a query param try getting an active contest instead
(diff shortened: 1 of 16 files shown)


Refresh.GameServer/Types/Levels/Categories/ContestCategory.cs
```diff
@@ -0,0 +1,30 @@
...
+    
+    public override DatabaseList<GameLevel>? Fetch(RequestContext context, int skip, int count, MatchService matchService,
+        GameDatabaseContext database, GameUser? user, LevelFilterSettings levelFilterSettings)
+    {
+        string? contestId = context.QueryString["contest"];
+        GameContest? contest = database.GetContestById(contestId);
+        if (contest == null) return null;
+        
+        return database.GetLevelsFromContest(contest, count, skip, user, levelFilterSettings);
+    }
```
Beyley, line 29: imo this should default to the latest contest (returning null when there is no active contest), so and also be unhidden and added to ingame categories, so for LBP3 where we can do custom level categories, the current latest contest will display as a level category.

Also did you forget to add this category to [here](https://github.com/LittleBigRefresh/Refresh/blob/4035585e12ca8f0190787c99dc16f57f08c6f424/Refresh.GameServer/Types/Levels/Categories/CategoryService.cs#L12)?

Beyley (changes requested):

Refresh.GameServer/Types/Levels/Categories/ContestCategory.cs
jvyden: I did forget, yes, thank you.

Good points, though, I'll change things to match that. I made the category pretty early on and then forgot about it so I haven't considered much about it.

Beyley (approved):

PR #462, opened 2024-05-06
Rewrite asset detection + add many more assets to our detection.

### LINKED ISSUE #438: Allow LBP3 Quests
Levels with Quests in them cannot be published (presumably due to missing assets)

Beyley:
This adds all missing assets from toolkit (thanks Aidan!)

We also now store the serialization method (binary, compressed texture, text, etc), and the "Asset Safety Level" differently.

Asset Safety Level has been replaced with a new AssetFlags flags enum, which stores whether a specific asset is "Dangerous", "Media", and/or "Modded".

This allows us to much more gradually block assets, e.g. block only Media assets, but allow other types of Modded assets, which was not possible before.

Opening as draft as i have yet to have the time to fully test this ingame, but tests pass locally

May fix #438 aswell, but that needs to be tested


Refresh.GameServer/Configuration/GameServerConfig.cs
```diff
@@ -9,16 +9,16 @@ namespace Refresh.GameServer.Configuration;
...
 
     protected override void Migrate(int oldVer, dynamic oldConfig) {}
 
     public string LicenseText { get; set; } = "Welcome to Refresh!";
 
-    public AssetSafetyLevel MaximumAssetSafetyLevel { get; set; } = AssetSafetyLevel.SafeMedia;
+    public AssetFlags BlockedAssetFlags { get; set; } = AssetFlags.Dangerous | AssetFlags.Modded;
     /// <seealso cref="GameUserRole.Trusted"/>
-    public AssetSafetyLevel MaximumAssetSafetyLevelForTrustedUsers { get; set; } = AssetSafetyLevel.SafeMedia;
+    public AssetFlags BlockedAssetFlagsForTrustedUsers { get; set; } = AssetFlags.Dangerous | AssetFlags.Modded;
```
jvyden, line 21: These need to have some sort of migration step.

Additionally we should provide documentation on setting this value, since I don't imagine bitwise values are easy to work with from a user's perspective. Maybe a class with boolean flags that get converted to the bitwise value instead?

Refresh.GameServer/Endpoints/ApiV3/DataTypes/Response/ApiInstanceResponse.cs
```diff
@@ -39,8 +39,8 @@ public class ApiInstanceResponse : IApiResponse
     public required string SoftwareLicenseUrl { get; set; }
     
     public required bool RegistrationEnabled { get; set; }
-    public required AssetSafetyLevel MaximumAssetSafetyLevel { get; set; }
-    public required AssetSafetyLevel MaximumAssetSafetyLevelForTrustedUsers { get; set; }
+    public required AssetFlags BlockedAssetFlags { get; set; } 
+    public required AssetFlags BlockedAssetFlagsForTrustedUsers { get; set; }
```
jvyden, line 43:
```suggestion
    public required AssetFlags BlockedAssetFlags { get; set; }
    public required AssetFlags BlockedAssetFlagsForTrustedUsers { get; set; }
```

Refresh.GameServer/Endpoints/Game/ResourceEndpoints.cs
```diff
@@ -67,21 +67,21 @@ public class ResourceEndpoints : EndpointGroup
...
-        // if asset safety is Safe (0), and maximum is configured as Safe (0), proceed
-        if (gameAsset.SafetyLevel > safetyLevel && !isPSP)
+        // Don't block any assets uploaded from PSP, else block any unwanted assets,
+        // For example, if the "blocked asset flags" has the "Media" bit set, and so does the asset,
+        // then that bit will be set after the AND operation, and we know to block it.
+        if ((gameAsset.AssetFlags & blockedAssetFlags) != 0 && !isPSP)
         {
             context.Logger.LogWarning(BunkumCategory.UserContent, $"{gameAsset.AssetType} {hash} by {user} is above configured safety limit " +
-                                                                 $"({gameAsset.SafetyLevel} > {safetyLevel})");
+                                                                 $"({gameAsset.AssetFlags} is blocked by {blockedAssetFlags})");
```
jvyden, line 80: This probably won't log the way we'd want it.

Does this log like `Flag1 | Flag2` or does it just spit out the raw number?

Refresh.GameServer/Importing/AssetImporter.cs
```diff
@@ -99,13 +99,16 @@ static char GetHexChar(int value)
             this.Warn($"{hash} is actually hashed as {checkedHash} - this asset is likely corrupt.");
             return null;
         }
-
+        
+        (GameAssetType gameAssetType, GameSerializationMethod gameSerializationMethod) = this.DetermineAssetType(data, platform);
```
jvyden, line 103: Should remove `game` prefix from variable names

Refresh.GameServer/Importing/Importer.cs
```diff
@@ -181,39 +181,29 @@ private bool IsMip(Span<byte> rawData)
...
-        if (MatchesMagic(data, "PRFb"u8)) return GameAssetType.SyncedProfile;
-        if (MatchesMagic(data, "MATT"u8)) return GameAssetType.GriefSongState;
-        if (MatchesMagic(data, "SSPt"u8)) return GameAssetType.SoftPhysicsSettings;
-        if (MatchesMagic(data, "BEVb"u8)) return GameAssetType.Bevel;
-        if (MatchesMagic(data, "ANMb"u8)) return GameAssetType.Animation;
+        GameAssetType? lbpAssetType = GameAssetTypeExtensions.LbpMagicToGameAssetType(data.Slice(0, 3));
+        
+        if (lbpAssetType != null)
+        {
+            return (lbpAssetType.Value, (GameSerializationMethod)data[3]);
```
jvyden, line 190: This cast to `GameSerializationMethod` doesn't feel safe?

Refresh.GameServer/Types/Assets/AssetFlags.cs
```diff
@@ -0,0 +1,114 @@
...
+    {
+        return type switch
+        {
+            // Non-binary levels will not be loadable 
+            GameAssetType.Level => method == GameSerializationMethod.Binary
+                ? AssetFlags.None
+                : AssetFlags.Dangerous | AssetFlags.Modded,
+            GameAssetType.StreamingLevelChunk => method == GameSerializationMethod.Binary
+                ? AssetFlags.None
+                : AssetFlags.Dangerous | AssetFlags.Modded,
```
jvyden, line 33: This is annoying to read... why are some of these values split into multiple while others are not?

I'd also statically import GameSerializationMethod and/or AssetFlags to make this slightly easier to read.

Refresh.GameServer/Types/Assets/AssetFlags.cs
```diff
@@ -0,0 +1,114 @@
...
+            GameAssetType.Plan => method == GameSerializationMethod.Binary ? AssetFlags.None : AssetFlags.Modded,
```
jvyden, line 34: Unsure about this "if not binary, then modded" classification. I suppose it's true that non-binary files being uploaded usually means an untrustworthy origin but if the game deserializes it the same, it should be treated the same. Not to mention code quality.

Refresh.GameServer/Types/Assets/AssetFlags.cs
```diff
@@ -0,0 +1,114 @@
...
+            GameAssetType.SoftPhysicsSettings => AssetFlags.Modded,
+            GameAssetType.Bevel => AssetFlags.Modded,
+            GameAssetType.Animation => AssetFlags.Modded,
+            GameAssetType.SettingsCharacter => AssetFlags.Modded,
+            GameAssetType.Joint => AssetFlags.Modded,
+            GameAssetType.MusicSettings => AssetFlags.Modded,
+            GameAssetType.StaticMesh => AssetFlags.Modded,
+            GameAssetType.PaintBrush => AssetFlags.Modded,
+            
+            GameAssetType.Script => AssetFlags.Dangerous | AssetFlags.Modded,
```
jvyden, line 74: Are there any cases of `Dangerous` without `Modded`? Makes the bitwise system in general feel a bit redundant.

Refresh.GameServer/Types/Assets/GameAssetType.cs
```diff
@@ -6,156 +7,654 @@ namespace Refresh.GameServer.Types.Assets;
...
     /// <summary>
-    /// An object, usually for prize bubbles or emitters and such.
+    /// A type of texture generally stored only in the game files, this is only ever uploaded when mods are used.
     /// </summary>
     /// <remarks>
-    /// Magic: PLNb
+    /// Magic: "GTF"
     /// </remarks>
-    Plan = 1,
+    GameDataTexture,
```
jvyden, line 25: Are the order changes here accounted for elsewhere? Have you tested migrations?

Refresh.GameServer/Types/Assets/GameAssetType.cs
```diff
@@ -6,156 +7,654 @@ namespace Refresh.GameServer.Types.Assets;
...
     /// </summary>
-    GriefSongState = 16,
+    /// <remarks>
+    /// Magic: "PAL"
+    /// </remarks>
+    Palette,
+    
+    /// <summary>
+    /// TODO
+    /// </summary>
```
jvyden, line 310: I'd rather the summaries just be removed instead of them being "TODO"

Refresh.GameServer/Types/Assets/GameAssetType.cs
```diff
@@ -6,156 +7,654 @@ namespace Refresh.GameServer.Types.Assets;
...
+    };
+    
+    public static GameAssetType? LbpMagicToGameAssetType(ReadOnlySpan<byte> magic)
+    {
+        int magicInt = magic[0] << 24 | magic[1] << 16 | magic[2] << 8;
+        
+        if (_lbpMagics.TryGetValue(magicInt, out GameAssetType assetType))
+            return assetType;
+        
+        return null;
```
jvyden, line 592:
```suggestion
        if (!_lbpMagics.TryGetValue(magicInt, out GameAssetType assetType))
            return null;
        
        return assetType;
```

Refresh.GameServer/Types/Assets/GameSerializationMethod.cs
```diff
@@ -0,0 +1,12 @@
+namespace Refresh.GameServer.Types.Assets;
+
+public enum GameSerializationMethod : byte
```
jvyden, line 3: This needs to be renamed to something shorter. Everywhere it's used causes readability problems.

Probably something like `GameDataFormat` or `GameAssetFormat`

jvyden (changes requested):
Feeling a bit mixed on the implementation of this.

jvyden:
Conflicting with master.

Also, are there any unit tests that should be added to help with this/asset detection in general? We have a few asset tests already but I'm wondering if we can do more.

Refresh.GameServer/Endpoints/Game/ResourceEndpoints.cs
Beyley: It does log how we want, it logs it like `(Flag1 | Flag2)`

Refresh.GameServer/Importing/Importer.cs
Beyley: It is safe, the table above only stores assets which follow LBP's standard serialization, and all assets which do that always will all have one of a known set of serialization methods

jvyden: Mostly worried about forged data. For example we might check against `Unknown` to see if it's valid, but the real value might be something that isn't in the bounds of the enum.

I would feel safer with a unit test proving this. Otherwise we should check against `Enum.IsDefined`.

Refresh.GameServer/Types/Assets/AssetFlags.cs
Beyley: that was something i was unsure about aswell, it depends on how precise we want to classify modded assets (since we want to provide a "is modded" flag for levels)

Refresh.GameServer/Types/Assets/GameSerializationMethod.cs
Beyley: i like `GameAssetFormat`, i'll go with that

Refresh.GameServer/Types/Assets/AssetFlags.cs
Beyley: > why are some of these values split into multiple while others are not?

I havent fully set the classification for all of them yet, but basically all asset type are "modded if not binary" (aside from textures ofc)

In this case, if we drop the "modded if not binary", then it will just be "dangerous if not binary", since the game is incapable of loading these assets when serialized in text form, and will cause an error for users if uploaded

Refresh.GameServer/Types/Assets/AssetFlags.cs
jvyden: Discard this; meant to delete

Refresh.GameServer/Types/Assets/AssetFlags.cs
jvyden: Meant to put the word lines there; this is a readability concern.

Some of these values are split into multiple lines. Most values are not; I don't see a reason for the inconsistency

Beyley: > Meant to put the word lines there; this is a readability concern.
> 
> Some of these values are split into multiple lines. Most values are not; I don't see a reason for the inconsistency

sometimes rider would format it one way, sometimes it wouldn't, i need to run rider's code formatter over this file before merge

Refresh.GameServer/Types/Assets/AssetFlags.cs
jvyden: I'm leaning towards the slightly less complicated solution of just treating all the types as the same, unless we can somehow make this less annoying to follow/work with.

Beyley: > I'm leaning towards the slightly less complicated solution of just treating all the types as the same, unless we can somehow make this less annoying to follow/work with.

I'm fine with treating them all the same, with the exception of unloadable ones (like `LVLt` or `TEXt`), those should definitely specified as "dangerous" in my opinion

perhaps it could be refactored to something like
```
AssetFlags flags = type switch {
	// All these branches are changed to ignore the serialization method
	GameAssetType.Level => AssetFlags.None,
};

//After we get the initial flags, we append on more flags for all of the unloadable ones, like `LVLt`, or `TXTt`
flags |= type switch {
	GameAssetType.Level => method == Binary ? AssetFlags.None : AssetFlags.Dangerous,
	GameAssetType.Texture => method == CompressedTexture ? AssetFlags.None : AssetFlags.Dangerous,
	else => AssetFlags.None,
};

return flags;
```

jvyden: Yeah, I prefer that actually.

Beyley:
> are there any unit tests that should be added to help with this/asset detection in general? We have a few asset tests already but I'm wondering if we can do more.

ideally we have a set of test assets stored in tree which test our dependency tree reading, asset type, serialization method, etc

Refresh.GameServer/Types/Assets/GameAssetType.cs
Beyley: I have tested migrations on my local instance, but i havent done it on a copy of prod yet, i'll do that before marking as ready for review

Beyley: double checked migrations, all works

Beyley:
No tests yet, but i resolved all the previous review comments

jvyden (changes requested):

Refresh.GameServer/Endpoints/ApiV3/DataTypes/Response/ApiInstanceResponse.cs
```diff
@@ -39,8 +39,8 @@ public class ApiInstanceResponse : IApiResponse
     public required string SoftwareLicenseUrl { get; set; }
     
     public required bool RegistrationEnabled { get; set; }
-    public required AssetSafetyLevel MaximumAssetSafetyLevel { get; set; }
-    public required AssetSafetyLevel MaximumAssetSafetyLevelForTrustedUsers { get; set; }
+    public required AssetFlags BlockedAssetFlags { get; set; }
+    public required AssetFlags BlockedAssetFlagsForTrustedUsers { get; set; }
```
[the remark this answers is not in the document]
Beyley: i think this is the same argument as "values vs names" for enums, and we ended up landing on values there 
imo since this is an api response, it makes more sense to present it as a bitflag number, since that will save bandwidth and mean that if we change the names Refresh-side, it doesnt break the API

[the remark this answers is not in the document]
jvyden: Changing the names is a massive refactor that simply isn't going to happen. If the names do get changed/shuffled, it would have caused breaking changes both internally and in the config files anyways.

The bandwidth overhead is not going to matter.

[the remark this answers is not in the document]
Beyley: thats fair, i'll change it to be configassetflags

Beyley:
still needs asset tests, but rebased to latest master and ready for more review

Refresh.GameServer/Importing/AssetImporter.cs
```diff
@@ -129,43 +132,30 @@ static char GetHexChar(int value)
...
+    private List<string> ParseAssetDependencies(byte[] data)
     {
-        List<string> dependencies = new();
-        
-        // Parse dependency table
         MemoryStream ms = new(data);
         BEBinaryReader reader = new(ms);
+        
+        // Read magic from the asset
+        reader.ReadUInt32();
```
jvyden, line 142:
```suggestion
        // Skip magic
        ms.Seek(4, SeekOrigin.Begin);
```

jvyden:
Looks fine but I probably won't review again until we have good tests for this.

jvyden:
This is bitrotting to hell and back so I'm just going to re-review, test it manually by re-importing all the assets locally and merge as-is if successful.

jvyden (approved):

jvyden:
Migration seems to be okay

PR #463, opened 2024-05-06
Contest improvements

turecross321:
This PR adds the following to contests:
- Theme
- Allowed games
- Template level


jvyden (approved):
looks like it won't set things on fire

PR #478, opened 2024-05-12
Implement rooms for LBP1 clients

Beyley:


Refresh.GameServer/Endpoints/Game/Handshake/AuthenticationEndpoints.cs
```diff
@@ -261,18 +262,21 @@ private static void SendPlatformNotAllowedNotification(GameDatabaseContext datab
...
     [GameEndpoint("goodbye", HttpMethods.Post, ContentType.Xml)]
     [MinimumRole(GameUserRole.Restricted)]
-    public Response RevokeThisToken(RequestContext context, GameDatabaseContext database, GameUser user)
+    public Response RevokeThisToken(RequestContext context, GameDatabaseContext database, GameUser user, Token token, MatchService matchService)
     {
-        string? token = context.Cookies["MM_AUTH"];
+        // On LBP1, our concept of a "room" is very basic, and we can be sure that they are the only user in said "room",
+        // so we just remove it outright if the user is logging out
+        // NOTE: We don't need any extra logic to handle more complex rooms here,
+        //       since LBP2 and LBP3 do not actually send requests to this endpoint when closed.
```
jvyden, line 270: I think they can, but the games usually segfault before the game can actually send this request. Best to not leave this as a special case. Only condition should be if the user is the host of that room.

Refresh.GameServer/Endpoints/Game/MatchingEndpoints.cs
```diff
@@ -37,4 +37,33 @@ public Response Match(RequestContext context, GameDatabaseContext database, Game
...
+            room.LevelType = RoomSlotType.Pod;
+            room.LevelId = id;
+        }
+        else
+        {
+            room.LevelType = slotType == "user" ? RoomSlotType.Online : RoomSlotType.Story;
+            room.LevelId = id;
+        }
+        
+        room.LastContact = DateTimeOffset.Now;
```
jvyden, line 63: Should probably make all requests update this field on LBP1. Or maybe just `/notification`; the only ticking request LBP1 makes.

Refresh.GameServer/Endpoints/Game/MatchingEndpoints.cs
```diff
@@ -37,4 +37,33 @@ public Response Match(RequestContext context, GameDatabaseContext database, Game
...
     }
+    
+    // Sent by LBP1 to notify the server it has entered a level.
+    // While it doesn't send us any detailed room information, we can at least create a pseudo-room on the server with just that player.
+    // Due to it only sending this when the level is *entered*, it means its much more likely for the room to be auto cleared due to inactivity,
+    // since the "bump" is much less often. This will at the very least make API tools be able to see LBP1 player activity and player counts
+    [GameEndpoint("enterLevel/{slotType}/{id}", HttpMethods.Post)]
+    public Response EnterLevel(RequestContext context, Token token, MatchService matchService, string slotType, int id)
+    {
+        GameRoom room = matchService.GetOrCreateRoomByPlayer(token.User, token.TokenPlatform, token.TokenGame, NatType.Moderate);
```
jvyden, line 48: Should assume worst-case scenario
```suggestion
        GameRoom room = matchService.GetOrCreateRoomByPlayer(token.User, token.TokenPlatform, token.TokenGame, NatType.Strict);
```

Refresh.GameServer/Endpoints/Game/Handshake/AuthenticationEndpoints.cs
```diff
@@ -261,18 +262,21 @@ private static void SendPlatformNotAllowedNotification(GameDatabaseContext datab
     /// </summary>
     [GameEndpoint("goodbye", HttpMethods.Post, ContentType.Xml)]
     [MinimumRole(GameUserRole.Restricted)]
-    public Response RevokeThisToken(RequestContext context, GameDatabaseContext database, GameUser user)
+    public Response RevokeThisToken(RequestContext context, GameDatabaseContext database, GameUser user, Token token, MatchService matchService)
     {
-        string? token = context.Cookies["MM_AUTH"];
+        // On LBP1, our concept of a "room" is very basic, and we can be sure that they are the only user in said "room",
+        // so we just remove it outright if the user is logging out
```
jvyden, line 268: Comments probably don't need to exist.

jvyden (changes requested):

Refresh.GameServer/Endpoints/Game/MatchingEndpoints.cs
Beyley: `/notification` is a good idea, i'll do that

Refresh.GameServer/Endpoints/Game/Handshake/AuthenticationEndpoints.cs
Beyley: Even when the games cleanly exited (tried on both PS3 and RPCS3, i could get the games to safely close) i couldn't get either to send this request, but I'll extend the logic to handle it anyway

jvyden: I'd prefer it be correct even if the games are stupid.

Refresh.GameServer/Endpoints/Game/Handshake/AuthenticationEndpoints.cs
```diff
@@ -262,23 +262,17 @@ private static void SendPlatformNotAllowedNotification(GameDatabaseContext datab
     /// </summary>
     [GameEndpoint("goodbye", HttpMethods.Post, ContentType.Xml)]
     [MinimumRole(GameUserRole.Restricted)]
-    public Response RevokeThisToken(RequestContext context, GameDatabaseContext database, GameUser user, Token token, MatchService matchService)
+    public Response RevokeThisToken(RequestContext context, GameDatabaseContext database, Token token, MatchService matchService)
```
jvyden, line 265: Keep the reference to the user; removing it is bound to cause issues when we revoke the token in this method.

jvyden (changes requested):

Refresh.GameServer/Endpoints/Game/Handshake/AuthenticationEndpoints.cs
Beyley: Fixed in #79f24aab24d93bf4a03729004fd2cbf27b22203d

jvyden (approved):

PR #486, opened 2024-05-12
Tweaks to object expiry

### LINKED ISSUE #418: Old tokens do not get removed from the database correctly
Prod has tokens from mid 2023

jvyden:
- [Make registrations expire in 1 hour](https://github.com/LittleBigRefresh/Refresh/commit/75a777670f85ddbff1fd668d1d8906c23b1a98c7)
    Hopefully shouldn't backfire. The goal is to help users who accidentally signed up with the wrong usernames. Should result in less admin intervention.
- [Remove tokens when expired](https://github.com/LittleBigRefresh/Refresh/commit/5d1bdab6fc22b88973014dc258ff7ca4f5b29aaf)
    We were originally storing these for historical purposes, but we never used this data.
    The first time this runs, it will take a very long time (roughly 10-30 minutes) on production as we don't do any batching when removing these tokens from Realm, but it won't be an issue as new tokens expire. **This will not block startup. This operation will run in the background.**

    Closes #418.
- [Shorten ExpiredObjectWorker work interval to 1 minute](https://github.com/LittleBigRefresh/Refresh/commit/d978f3a00c52511e0948e8a21e1b0295d246d1a8)
    Don't see why not.

### DIFF
--- a/Refresh.GameServer/Endpoints/ApiV3/AuthenticationApiEndpoints.cs
+++ b/Refresh.GameServer/Endpoints/ApiV3/AuthenticationApiEndpoints.cs
@@ -263,8 +263,8 @@ public class AuthenticationApiEndpoints : EndpointGroup
             database.AddRegistrationToQueue(body.Username, body.EmailAddress, passwordBcrypt);
             return new ApiAuthenticationError(
                 "Your account has been put into the registration queue, but it is not yet activated. " +
-                "To complete registration, simply log in from LBP within the next hour and your new account will be activated. " +
-                "You will be unable to sign in until you are patched and playing.", true);
+                "To complete registration, patch your games to our servers and start playing within the next hour and your new account will be activated. " +
+                "You will be unable to sign in until you are patched and playing. For more instructions on patching, please visit https://docs.littlebigrefresh.com", true);
         }
 
         GameUser user = database.CreateUser(body.Username, body.EmailAddress, true);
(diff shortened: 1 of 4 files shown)


Refresh.GameServer/Endpoints/ApiV3/AuthenticationApiEndpoints.cs
```diff
@@ -262,8 +262,9 @@ public ApiOkResponse DenyVerificationRequest(RequestContext context, GameDatabas
         {
             database.AddRegistrationToQueue(body.Username, body.EmailAddress, passwordBcrypt);
             return new ApiAuthenticationError(
-                "Your account has been created, but it is not yet activated. " +
-                "To complete registration, simply log in from LBP and your new account will be activated.", true);
+                "Your account has been put into the registration queue, but it is not yet activated. " +
+                "To complete registration, simply log in from LBP within the next hour and your new account will be activated. " +
```
Beyley, line 266: Something ive learned helping people connect, they dont understand the phrase "log in", perhaps it should be _something_ along the lines of "To complete registration, patch LBP, and then start browsing community levels. or your registration will be deleted in 1 hour", where it explains that "log in" means "start browsing community levels/access ingame online features". the words "login" isnt really used by the game to explain the connection to end users

jvyden: image
How's this?

Beyley: should be much better

Beyley (approved):

PR #488, opened 2024-05-13
Move request tracking submission to background worker

jvyden:
This massively improves performance of ***all*** requests by incrementing the request counters in a background thread every 5 seconds, as opposed to the previous solution of incrementing by one upon every request.

Since we are no longer committing to Realm every request, we also remove the ability to disable request tracking as there is no performance penalty for having it on.

This brings all requests down from roughly an average of 8-16ms all the way down to 1-2ms. For large swarms of requests such as a level downloader grabbing assets from us at high speeds, this can reduce write speeds of up to 30mb/s to practically nothing, as well as reducing peak request times from 130ms to the same 1-2ms.

### DIFF
--- a/Refresh.GameServer/Services/RequestStatisticTrackingService.cs
+++ b/Refresh.GameServer/Services/RequestStatisticTrackingService.cs
@@ -4,22 +4,19 @@ using Bunkum.Core.Database;
 using Bunkum.Core.Responses;
 using Bunkum.Core.Services;
 using NotEnoughLogs;
-using Refresh.GameServer.Database;
 using Refresh.GameServer.Endpoints;
 
 namespace Refresh.GameServer.Services;
 
 public class RequestStatisticTrackingService : Service
 {
-    public static readonly object TrackerLock = new();
+    private static int _gameRequestsToSubmit;
+    private static int _apiRequestsToSubmit;
     
-    public static int GameRequestsToSubmit { get; private set; }
-    public static int ApiRequestsToSubmit { get; private set; }
-    
-    public static void ClearRequests()
+    public static (int game, int api) SubmitAndClearRequests()
     {
-        GameRequestsToSubmit = 0;
-        ApiRequestsToSubmit = 0;
+        return (game: Interlocked.Exchange(ref _gameRequestsToSubmit, 0),
+                api: Interlocked.Exchange(ref _apiRequestsToSubmit, 0));
     }
     
     internal RequestStatisticTrackingService(Logger logger) : base(logger)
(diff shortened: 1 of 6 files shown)


Refresh.GameServer/Workers/RequestStatisticSubmitWorker.cs
```diff
@@ -0,0 +1,24 @@
...
+    
+    public void DoWork(Logger logger, IDataStore dataStore, GameDatabaseContext database)
+    {
+        lock (TrackingService.TrackerLock)
+        {
+            database.IncrementGameRequests(TrackingService.GameRequestsToSubmit);
+            database.IncrementApiRequests(TrackingService.ApiRequestsToSubmit);
+            
+            TrackingService.ClearRequests();
+        }
```
Beyley, line 22: This can be done lockless with [`Increment.Exchange`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.interlocked.exchange?view=net-8.0#system-threading-interlocked-exchange(system-int32@-system-int32))

Just exchange the value with 0, and pass the returned value into the increment calls

Refresh.GameServer/Services/RequestStatisticTrackingService.cs
```diff
@@ -11,20 +11,32 @@ namespace Refresh.GameServer.Services;
...
-            gameDatabase.IncrementApiRequests();
+            if (context.Uri.AbsolutePath.StartsWith(GameEndpointAttribute.BaseRoute))
+            {
+                GameRequestsToSubmit++;
+            }
+            else
+            {
+                ApiRequestsToSubmit++;
+            }
         }
```
Beyley, line 40: This can be done lock-less using [`Interlocked.Increment`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.interlocked.increment?view=net-8.0#system-threading-interlocked-increment(system-int32@))

Beyley (changes requested):
this code can be made much more efficient by removing the need for locking

Refresh.GameServer/Services/RequestStatisticTrackingService.cs
```diff
@@ -4,28 +4,30 @@
...
         if (context.Uri.AbsolutePath.StartsWith(GameEndpointAttribute.BaseRoute))
-        {
-            gameDatabase.IncrementGameRequests();
-        }
+            Interlocked.Exchange(ref _gameRequestsToSubmit, _gameRequestsToSubmit + 1);
         else
-        {
-            gameDatabase.IncrementApiRequests();
-        }
+            Interlocked.Exchange(ref _apiRequestsToSubmit, _apiRequestsToSubmit + 1);
```
Beyley, line 30: These should be using `Interlocked.Increment(ref value)`
this will be compiled down to the operations
- read i32 into register (this part is atomic)
- increment register
- exchange value with the incremented register (this part is atomic)

since the whole operation is not done atomically, theres the chance for race conditions

while interlocked.increment will be compiled down to the operation
- increment value at address (atomic)

Beyley (changes requested):

Refresh.GameServer/Services/RequestStatisticTrackingService.cs
jvyden: oh. yeah, i did not read the second comment properly.

Beyley (approved):

PR #495, opened 2024-05-14
Replace `FillInExtraData` system with a new `DataContext` object in `IDataConvertableFrom`

jvyden:
This PR modifies a significant chunk of the codebase for endpoints, so for an easier review I recommend just looking at only the files regarding the `DataContext` itself and the service for it.

Key things to consider:
- A change to `GameDatabaseContext.GetAssetFromHash` to return null for assets that will never be stored in the database. Simplifies some code and makes it faster.
- Since the `DataContext` service depends on most services, it must be added as the last service, always.
- `DataContext.Game` and `DataContext.Platform` both assume `Website` if a token is not present. This shouldn't ever be hit in Game API code but it's something to be aware of.
- To simplify code when making a new `DataContext`, `DataContextService.StealInjection` does some illegal things by passing `null` for some parameters. This should be okay for the services that are already there, but I recommend looking at the services `AddParameterToEndpoint` function when doing this just to make sure.
- Most of this refactor was done automatically with Rider's refactoring tools. Unused parameters in endpoints are not considered, and ordering of parameters in endpoints are also not considered. I'm not willing to manually give each little endpoint a correction of this, because it doesn't matter in the long run.

### DIFF
--- a/Refresh.GameServer/Database/DatabaseList.cs
+++ b/Refresh.GameServer/Database/DatabaseList.cs
@@ -1,4 +1,5 @@
 using Refresh.GameServer.Endpoints.ApiV3.DataTypes;
+using Refresh.GameServer.Types.Data;
 
 namespace Refresh.GameServer.Database;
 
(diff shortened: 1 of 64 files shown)


Beyley (approved):
absolutely loving this refactor even more now that i see it in code, finally one of the biggest warts of the codebase is gone

PR #517, opened 2024-05-29
Implement various API extensions

### LINKED ISSUE #515: API Extensions
The redesign has brought up a lot of interesting discussions, but unfortunately a lot of the data shown in our designs isn't available from the API.

Here's what we should definitely be serializing:

## Users
- [x] Role
- [x] Active room *(this can technically be gathered already, but should be bundled for easier access)*
- [x] Statistics
    - [x] Heart count
    - [x] Total comments on profile
    - [x] Published level count
    - [x] Taken photos count
## Extended Users
- [x] File size quota usage
## Comments
Comments, whether on a level or user, are entirely inaccessible from the API.
## Levels
- [x] Statistics
    - [x] Photos taken in level
    - [x] Total comments on level
    - [x] Total reviews on level

jvyden:
Closes #515.

Ideally I would have included level statistics in a new Statistics object like I've done for users for organization purposes, but there are already statistics present in ApiGameLevelResponse and I don't want to have them split up.

Something to consider for APIv4 if we ever consider jumping from v3.

### DIFF
--- a/Refresh.GameServer/Database/GameDatabaseContext.Comments.cs
+++ b/Refresh.GameServer/Database/GameDatabaseContext.Comments.cs
@@ -27,6 +27,8 @@ public partial class GameDatabaseContext // Comments
             .AsEnumerable()
             .Skip(skip)
             .Take(count);
+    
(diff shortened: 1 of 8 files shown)


Beyley (approved):
pona

PR #519, opened 2024-06-04
Matching + Metadata fixes

### LINKED ISSUE #373: Deprioritize rooms with `PassedNoJoinPoint` set when matching
When the game updates the room state vita `/lbp/match`, it tells us whether the player has passed a no join gate, we should de-prioritize these rooms in the matching process, so that players are more likely to join games which they will be able to immediately play in.

Beyley:
Easiest to review by commit

The metadata commit implements the start to the "level/profile visibility" functionality the game exposes, but lacks the actual functionality of hiding the user/levels from the API 

This also adds multiple fixes to matching, making vita dive in work, and implements "passed no join point" 

Closes #373


jvyden (changes requested):
these need to be two different PRs

PR #520, opened 2024-06-04
Metadata: Implement storage/retrieval of level/profile visibility

Beyley:
This does not implement the functionality, but it means we are now storing all the relevant information to make that happen.

### DIFF
--- a/Refresh.GameServer/Database/GameDatabaseContext.Users.cs
+++ b/Refresh.GameServer/Database/GameDatabaseContext.Users.cs
@@ -4,6 +4,7 @@ using MongoDB.Bson;
 using Realms;
 using Refresh.GameServer.Authentication;
 using Refresh.GameServer.Endpoints.ApiV3.DataTypes.Request;
+using Refresh.GameServer.Endpoints.Game.Handshake;
 using Refresh.GameServer.Types;
 using Refresh.GameServer.Types.Levels;
 using Refresh.GameServer.Types.Photos;
@@ -387,19 +388,15 @@ public partial class GameDatabaseContext // Users
         });
     }
     
-    public void SetProfileVisibility(GameUser user, Visibility visibility)
+    public void SetPrivacySettings(GameUser user, MetadataEndpoints.PrivacySettings settings) 
     {
         this._realm.Write(() =>
         {
-            user.ProfileVisibility = visibility;
-        });
-    }
-    
-    public void SetLevelVisibility(GameUser user, Visibility visibility)
-    {
-        this._realm.Write(() =>
-        {
-            user.LevelVisibility = visibility;
-        });
+            if(settings.LevelVisibility != null)
+                user.LevelVisibility = settings.LevelVisibility.Value;
+            
+            if (settings.ProfileVisibility != null)
+                user.ProfileVisibility = settings.ProfileVisibility.Value;
+        });            
     }
 }
--- a/Refresh.GameServer/Database/GameDatabaseContext.Users.cs
+++ b/Refresh.GameServer/Database/GameDatabaseContext.Users.cs
@@ -1,10 +1,7 @@
-using System.Reflection;
 using JetBrains.Annotations;
 using MongoDB.Bson;
-using Realms;
 using Refresh.GameServer.Authentication;
 using Refresh.GameServer.Endpoints.ApiV3.DataTypes.Request;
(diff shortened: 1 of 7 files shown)


Refresh.GameServer/Database/GameDatabaseContext.Users.cs
```diff
@@ -380,4 +386,20 @@ public void IncrementUserFilesizeQuota(GameUser user, int amount)
...
+        });
+    }
+    
+    public void SetLevelVisibility(GameUser user, Visibility visibility)
+    {
+        this._realm.Write(() =>
+        {
+            user.LevelVisibility = visibility;
+        });
+    }
```
jvyden, line 404: These parameters technically aren't bundled in one request from the game, but to simplify the endpoint and to keep this class shorter I think it should be one function called `SetUserPrivacySettings`.

Refresh.GameServer/Endpoints/ApiV3/DataTypes/Request/ApiUpdateUserRequest.cs
```diff
@@ -14,4 +16,7 @@ public class ApiUpdateUserRequest
     public bool? UnescapeXmlSequences { get; set; }
     
     public string? EmailAddress { get; set; }
+    
+    public Visibility? LevelVisibility { get; set; }
+    public Visibility? ProfileVisibility { get; set; }
```
jvyden, line 21: New properties in this class should be handled in the corresponding function in `GameDatabaseContext`. The function in question is the `ApiUpdateUserRequest` overload of `UpdateUserData`.

This makes it so these properties can actually be updated from the API - we removed the reflection-based updating a while ago.

Refresh.GameServer/Endpoints/Game/Handshake/MetadataEndpoints.cs
```diff
@@ -18,15 +18,31 @@ public class MetadataEndpoints : EndpointGroup
...
+        {
+            LevelVisibility = user.LevelVisibility,
+            ProfileVisibility = user.ProfileVisibility,
+        };
     }
     
     [GameEndpoint("privacySettings", ContentType.Xml, HttpMethods.Post)]
-    public PrivacySettings SetPrivacySettings(RequestContext context)
+    [MinimumRole(GameUserRole.Restricted)]
+    [DebugRequestBody]
```
jvyden, line 32: `[DebugRequestBody]` Should be nuked

jvyden (changes requested):

Refresh.GameServer/Endpoints/ApiV3/DataTypes/Request/ApiUpdateUserRequest.cs
Beyley: i did this [here](https://github.com/LittleBigRefresh/Refresh/pull/520/files#diff-82503085bd955b6efc6eb6353f5f08e8cb21ac466bb03eaad9d5b0e7f3a0eaf8R181)

Refresh.GameServer/Database/GameDatabaseContext.Users.cs
Beyley: so basically, in the implementation of the endpoint do `get current settings -> update changed fields -> update whole settings`?

jvyden: Just move the logic from the endpoint to a function in the database that applies based off of the privacy settings object. The implementation itself doesn't need to change, apart from putting everything in a write block I guess.

Refresh.GameServer/Database/GameDatabaseContext.Users.cs
```diff
@@ -380,4 +387,16 @@ public void IncrementUserFilesizeQuota(GameUser user, int amount)
             user.FilesizeQuotaUsage += amount;
         });
     }
+    
+    public void SetPrivacySettings(GameUser user, MetadataEndpoints.PrivacySettings settings)
```
jvyden, line 391: Ouch. Yeah, I would extract PrivacySettings to its own class. `SerializedPrivacySettings` somewhere in `Types.UserData`.

Generally speaking the database should never depend on endpoints, and ideally never depend on Bunkum or any services. Reason being I hope to someday move the database schema to its own project in the solution.

jvyden (changes requested):

jvyden (approved):

PR #521, opened 2024-06-04
Matching: Vita + PassedNoJoinPoint + Better slots handling

### LINKED ISSUE #373: Deprioritize rooms with `PassedNoJoinPoint` set when matching

Beyley:
Closes #373

### DIFF
--- a/Refresh.GameServer/Services/MatchService.cs
+++ b/Refresh.GameServer/Services/MatchService.cs
@@ -27,7 +27,7 @@ public partial class MatchService(Logger logger) : EndpointService(logger)
         TokenGame game, 
         NatType natType, 
         int? buildVersion,
-        bool passedNoJoinPoint)
+        bool? passedNoJoinPoint = null)
     {
         GameRoom? room = this.RoomAccessor.GetRoomByUser(player, platform, game);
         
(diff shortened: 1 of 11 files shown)


Refresh.GameServer/Services/MatchService.cs
```diff
@@ -21,38 +21,56 @@ public partial class MatchService(Logger logger) : EndpointService(logger)
...
-    public GameRoom GetOrCreateRoomByPlayer(GameUser player, TokenPlatform platform, TokenGame game, NatType natType)
+    public GameRoom GetOrCreateRoomByPlayer(
+        GameUser player, 
+        TokenPlatform platform, 
+        TokenGame game, 
+        NatType natType, 
+        int? buildVersion,
+        bool passedNoJoinPoint)
     {
         GameRoom? room = this.RoomAccessor.GetRoomByUser(player, platform, game);
```
jvyden, line 32: I see this being initialized to `false` a lot depending on a nullable value, which makes me think it should work like this:

```suggestion
        bool? passedNoJoinPoint = null)
    {
        passedNoJointPoint ??= false;
        GameRoom? room = this.RoomAccessor.GetRoomByUser(player, platform, game);
```

And then a consumer can just pass in the `Nullable<bool>` without a `?? false`.

jvyden (changes requested):

jvyden (approved):

PR #523, opened 2024-06-12
Assets: Make asset dependencies postgres friendly

Beyley:


Refresh.GameServer/Types/Assets/AssetDependencyRelation.cs
```diff
@@ -0,0 +1,9 @@
+using Realms;
+
+namespace Refresh.GameServer.Types.Assets;
+
+public partial class AssetDependencyRelation : IRealmObject
+{
+    public string Dependent { get; set; }
```
jvyden, line 7: should disable nullable in model classes unless necessary to avoid these compiler warnings

Refresh.GameServer/Refresh.GameServer.csproj
```diff
@@ -71,7 +71,7 @@
       <PackageReference Include="MailKit" Version="4.6.0" />
       <PackageReference Include="NPTicket" Version="3.1.0" />
       <PackageReference Include="Pfim" Version="0.11.2" />
-      <PackageReference Include="Realm" Version="11.7.0" />
+      <PackageReference Include="Realm" Version="12.2.0" />
```
jvyden, line 74: probably needs a bump bunkum-side as well.

Refresh.GameServer/Database/GameDatabaseContext.Assets.cs
```diff
@@ -30,6 +30,34 @@ public partial class GameDatabaseContext // AssetConfiguration
         
         return null;
     }
+
+    public IQueryable<AssetDependencyRelation> GetAssetDependencies(GameAsset asset) 
+        => this._realm.All<AssetDependencyRelation>().Where(a => a.Dependent == asset.AssetHash);
```
jvyden, line 35: the return type complicates thing for code that uses this... the dependent is already implied from the `asset` so i think you should just `.AsEnumerable().Select(x => x.Dependency)` and return that instead to get the old behavior back

jvyden (changes requested):

jvyden (approved):

PR #531, opened 2024-06-17
Remove `Refresh.Analyzers` project

### LINKED ISSUE #88: Rework Recent Activity source generator
It's an absolute mess, and it depends on a lot of hacks/assumptions to work correctly.

jvyden:
Closes #528 and closes #88.

`Refresh.Analyzers` has been causing issues since it was implemented. Whether it be `CodeAnalysis.CSharp` updating and breaking workflows, the challenge that comes with implementing new events, or causing hacky naming in `EventType`, it's never been good and probably never will be.

It was intended to alleviate copy+pasting when making new events but it's caused more pain than it's helped. So, let's just remove it.

To do this I copied the output of the generated classes, cleaned them up a bit to keep things more in-line with the rest of the codebase, moved the database partials for recent activity to their own subfolder, and then deleted the `Refresh.Analyzers` project entirely. I also took the opportunity to fix  the somewhat cursed naming of the `EventType` enum for consistency.


Refresh.GameServer/Types/Activity/EventType.cs
```diff
@@ -1,57 +1,55 @@
...
-    // Playlist_AddLevel = 19,
+    // PlaylistAddLevel = 19,
     [XmlEnum("score")]
-    SubmittedScore_Create = 20, // FIXME: this name is shit
+    LevelScore = 20,
     
     // Custom events
     [XmlEnum("firstlogin")]
-    User_FirstLogin = 127,
+    UserFirstLogin = 127,
```
Beyley, line 54: Is there a reason a lot of these are still commented?

Beyley (commented):
lgtm, but i do have a question

Refresh.GameServer/Types/Activity/EventType.cs
jvyden: Before, they would cause methods to generate and cause problems. That won't happen now so I'll uncomment them.

Beyley (approved):

PR #535, opened 2024-06-18
Remove `LevelPlay` events

jvyden:
I don't think we need to be tracking every single time every user has ever decided to play a level, especially given the fact that many players spam-retry levels. I think the best course of action is to only record actual score submissions.

This should also drastically reduce the size of the Realm, and thus migration times hopefully.


Beyley:
imo we should still create level play events, but only if its a unique play, eg. their first time playing the level
that fixes the spam issue, while still keeping the functionality of the play events relatively the same to end users
not all levels will cause score submissions

jvyden:
I think that'd still cause a lot of clutter for what's not really much information.

No user is depending on Recent Activity like this I don't think.

Beyley:
> No user is depending on Recent Activity like this I don't think.

i personally do actually like to go into to the recent activity page ingame and play levels from there, and not having any play relations cuts out any levels that my friends may have played but then not finished, or maybe the level just didnt have an ending which submits a score

see here, where the level was played, but not beaten, without this event, the level just wouldnt appear in recent activity as often
its possible arcadius had already played it so it wouldnt have appeared anyway, but not beating levels/not having a scoreboard isnt too uncommon of an occurance, especially with giveaways, where theres no scoreboard, so no scores are submitted
image


i like the idea of not spamming multiple playlevel events, but i think the first unique play should still show an event

jvyden:
I 100% disagree with `LevelPlay` not being the most common event, but I'll add unique events back.

Refresh.GameServer/Types/Activity/ActivityPage.cs
```diff
@@ -233,7 +233,6 @@ private void GenerateLevelGroups(ActivityGroups groups)
             levelEvent = @event.EventType switch
             {
                 EventType.LevelUpload => SerializedLevelUploadEvent.FromSerializedLevelEvent(levelEvent),
-                EventType.LevelPlay => SerializedLevelPlayEvent.FromSerializedLevelEvent(levelEvent),
```
Beyley, line 236: This and other code needs to be re-added for [this](https://github.com/LittleBigRefresh/Refresh/pull/535/files#diff-e94d1846e9d44e7d024aa109cc13112930742e5fc3b961fa1cd1c7f301fe7922R325) to work again

Beyley (approved):

PR #537, opened 2024-06-19
Prevent users from re-publishing a level twice.

Beyley:
This should curb the reuploads on the official instance by a small amount. There isnt a good reason for a person to want to do this in a circumstance other than spamming/uploading a duplicate level. This is something we'd want to do after the dry archive import anyway, so might as well do it now since its very little code, and helps an immediate issue.

### DIFF
--- a/Refresh.GameServer/Database/GameDatabaseContext.Levels.cs
+++ b/Refresh.GameServer/Database/GameDatabaseContext.Levels.cs
@@ -7,7 +7,6 @@ using Refresh.GameServer.Endpoints.ApiV3.DataTypes.Request;
 using Refresh.GameServer.Endpoints.Game.Levels.FilterSettings;
 using Refresh.GameServer.Extensions;
 using Refresh.GameServer.Services;
-using Refresh.GameServer.Types;
 using Refresh.GameServer.Types.Activity;
 using Refresh.GameServer.Types.Levels;
 using Refresh.GameServer.Types.Matching;
@@ -391,6 +390,9 @@ public partial class GameDatabaseContext // Levels
                 return null;
         }
     }
+
+    public GameLevel? GetLevelByRootResource(string rootResource) =>
+        this.GameLevels.FirstOrDefault(level => level.RootResource == rootResource);
     
     [Pure]
     public GameLevel? GetLevelById(int id) => this.GameLevels.FirstOrDefault(l => l.LevelId == id);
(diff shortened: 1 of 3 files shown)


Refresh.GameServer/Endpoints/Game/Levels/PublishEndpoints.cs
```diff
@@ -52,6 +52,15 @@ private static bool VerifyLevel(GameLevelRequest body, GameUser user, Logger log
             if (!guidChecker.IsTextureGuid(game, long.Parse(body.IconHash.AsSpan()[1..])))
                 return false;
         }
+
+        GameLevel? existingLevel = database.GetLevelByRootResource(body.RootResource);
+        // If there is an existing level with this root hash, and this isn't an update request, block the upload
+        if (existingLevel != null && body.LevelId != existingLevel.LevelId)
+        {
+            database.AddPublishFailNotification("You may not re-upload another user's level.", body.ToGameLevel(user), user);
```
jvyden, line 60:
```suggestion
            database.AddPublishFailNotification("The level you tried to publish has already been uploaded by another user.", body.ToGameLevel(user), user);
```

Refresh.GameServer/Endpoints/Game/Levels/PublishEndpoints.cs
```diff
@@ -29,7 +29,7 @@ public class PublishEndpoints : EndpointGroup
     /// <param name="guidChecker">The associated GuidCheckerService with the request</param>
     /// <param name="game">The game the level is being submitted from</param>
     /// <returns>Whether or not validation succeeded</returns>
-    private static bool VerifyLevel(GameLevelRequest body, GameUser user, Logger logger, GuidCheckerService guidChecker, TokenGame game)
+    private static bool VerifyLevel(GameLevelRequest body, GameUser user, Logger logger, GuidCheckerService guidChecker, TokenGame game, GameDatabaseContext database)
```
jvyden, line 32: This is getting to the point to where we should probably use a `DataContext`

jvyden (changes requested):

Refresh.GameServer/Database/GameDatabaseContext.Levels.cs
```diff
@@ -391,6 +390,9 @@ public int GetTotalLevelsPublishedByUser(GameUser user, TokenGame game)
                 return null;
         }
     }
+
+    public GameLevel? GetLevelByRootResource(string rootResource) =>
+        this.GameLevels.FirstOrDefault(level => level.RootResource == rootResource);
```
jvyden, line 395: Obviously out of scope but we should provide an API endpoint for this :)

`/api/v3/levels/hash/{hash}`

jvyden (approved):

PR #548, opened 2024-06-30
FindRoomMethod: Fix level ID search & always print room state

Beyley:
Printing the room state out always helps a lot with debugging when people state that "dive in doesn't work". 
Often by the time someone responds "can you take a capture of the API room output", the room state has changed enough where it's not as helpful, the API also doesn't expose info like NAT type or build version. This change makes sure that we can easily see exactly what the server was seeing when it filtered down the rooms, and can narrow down why no rooms were found.

If the extra noise does become a problem in the future when we grow and have many rooms, I have made the dump a config option, so that it can be easily disabled/re-enabled when needed.

The level ID search changes should also fix some issues people have been having with dive in.

### DIFF
--- a/Refresh.GameServer/Types/Matching/MatchMethods/IMatchMethod.cs
+++ b/Refresh.GameServer/Types/Matching/MatchMethods/IMatchMethod.cs
@@ -2,6 +2,7 @@ using Bunkum.Core.Responses;
 using JetBrains.Annotations;
 using NotEnoughLogs;
 using Refresh.GameServer.Authentication;
+using Refresh.GameServer.Configuration;
 using Refresh.GameServer.Database;
 using Refresh.GameServer.Services;
 using Refresh.GameServer.Types.UserData;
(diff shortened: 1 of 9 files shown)


Refresh.GameServer/Types/Matching/MatchMethods/IMatchMethod.cs
```diff
@@ -14,5 +15,6 @@ public interface IMatchMethod
     IEnumerable<string> MethodNames { get; }
 
     Response Execute(MatchService service, Logger logger,
-        GameDatabaseContext database, GameUser user, Token token, SerializedRoomData body);
+        GameDatabaseContext database, GameUser user, Token token, SerializedRoomData body,
+        GameServerConfig gameServerConfig);
 }
```
jvyden, line 20: should use `DataContext`

jvyden (changes requested):
looks fine, but `IMatchMethod` is getting messy:

Refresh.GameServer/Types/Matching/MatchMethods/IMatchMethod.cs
Beyley: switched to using a data context

jvyden (approved):

Issue #492, opened 2024-05-14
Curator role


jvyden:
Suggested on Discord by CrankyLandlord & StarflightGamin.

This role, when applied to a user, would grant access to the admin panel component of a level. This would allow them to team pick levels and set reupload data, and adjust the level's game version.

This doesn't seem too hard to implement. I'd probably set the new role between trusted and admin, but not after trusted. Probably `64` or something to put it between them with room for more roles:

```
    Admin = 127,

    Curator = 64,

    Trusted = 1,
    User = 0,
    Restricted = -126,
    Banned = -127,
```

StarflightGamin:
Would it be possible to make it so being given this role unlocks the MM employee pin? Just a fun little badge of honor that people who help team pick levels on Refresh can display, although I can understand it not happening lol.

Beyley:
> Would it be possible to make it so being given this role unlocks the MM employee pin? Just a fun little badge of honor that people who help team pick levels on Refresh can display, although I can understand it not happening lol.

If anything i would prefer only the Refresh team members (eg. those marked as "maintainer/staff" on the discord) get the MM employee pin, i like the idea of adding functionality back for some of the special pins, but i think it gives the wrong impression if its for a completely different purpose than the original pin's purpose

StarflightGamin:
Yea, that's fair lol.

PR #556, opened 2024-07-09
Refactor CreateUser to accept a role to set, implement curator role

jvyden:
Implemented as described in #492.

Includes a bonus refactor to `TestContext` to make making users with roles from within tests a bit easier.


Beyley (approved):
pona

PR #557, opened 2024-07-11
Add Refresh.HttpsProxy

Beyley:
This adds a simple proxy capable of forwarding LBP3 traffic on ports 10061 (HTTPS) and 10060 (HTTP) to another gameserver, it also overrides the original server's digest, allowing clients with non-patched digest keys to connect without the actual gameserver having to support the proxied game's digest.

### DIFF
--- a/Refresh.HttpsProxy/Middlewares/ProxyMiddleware.cs
+++ b/Refresh.HttpsProxy/Middlewares/ProxyMiddleware.cs
@@ -30,9 +30,7 @@ public class ProxyMiddleware(ProxyConfig config) : IMiddleware
         for (int i = 0; i < context.Query.Count; i++)
         {
             string key = context.Query.GetKey(i)!;
-            
             string name = HttpUtility.HtmlEncode(key);
-
             string value = HttpUtility.HtmlEncode(context.Query[key]!);
 
             queryString.AppendFormat("{0}={1}", name, value);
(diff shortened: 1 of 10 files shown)


Refresh.HttpsProxy/Middlewares/ProxyMiddleware.cs
```diff
@@ -0,0 +1,82 @@
...
+            queryString.Append('?');
+
+        // Fill in all the query strings
+        for (int i = 0; i < context.Query.Count; i++)
+        {
+            string key = context.Query.GetKey(i)!;
+            
+            string name = HttpUtility.HtmlEncode(key);
+
+            string value = HttpUtility.HtmlEncode(context.Query[key]!);
```
jvyden, line 36:
```suggestion
            string key = context.Query.GetKey(i)!;
            string name = HttpUtility.HtmlEncode(key);
            string value = HttpUtility.HtmlEncode(context.Query[key]!);
```

jvyden (changes requested):
Couple small things

Will you be moving the digest changes over to the current middleware implementation at all or are you just keeping the refactor to the new one?

Beyley:
> Will you be moving the digest changes over to the current middleware implementation at all or are you just keeping the refactor to the new one?

I will do that in a future PR

jvyden (approved):

jvyden:
docker support would be helpful but i'll address that later

PR #563, opened 2024-07-17
Implement rating reviews

### LINKED ISSUE #449: Rating Reviews
LittleBigPlanet allows you to rate comments and reviews, which refresh is currently missing. implementing this would be benefitial, since rating reviews places them higher or lower depending on the yay/boo ratio it gets

speedyman08:
Closes #449

### DIFF
--- a/Refresh.GameServer/Database/GameDatabaseContext.Relations.cs
+++ b/Refresh.GameServer/Database/GameDatabaseContext.Relations.cs
@@ -192,12 +192,32 @@ public partial class GameDatabaseContext // Relations
 
     #region Rating and Reviewing
 
-    public void RateReview(GameReview review, RatingType rt)
+    public void RateReview(GameReview review, RatingType rt, GameUser user)
     {
+        // if rating type is neutral, remove previous review rating by user 
+        if (rt == RatingType.Neutral && this.ReviewRatingExistsByUser(user, review))
+        {
+            RateReviewRelation rel =
+                this.RateReviewRelations.First(relation => relation.Review == review && relation.User == user);
+            this.Write(() => this.RateReviewRelations.Remove(rel));
+
+            return;
+        }
+        // swap rating
+        if (this.ReviewRatingExistsByUser(user, review))
+        {
+            RateReviewRelation rel = this.RateReviewRelations.First(relation => relation.Review == review && relation.User == user);
+            
+            this.Write(() => rel.RatingType = rt);
+
+            return;
+        }
+        
         RateReviewRelation reviewRelation = new()
         {
             RatingType = rt,
             Review = review,
+            User = user,
         };
 
         this.Write(() =>
@@ -205,7 +225,7 @@ public partial class GameDatabaseContext // Relations
             this.RateReviewRelations.Add(reviewRelation);
         });
     }
-
+    
     public ReviewRating GetRatingForReview(GameReview review)
     {
         IQueryable<RateReviewRelation> relations = this.RateReviewRelations.Where((relation => relation.Review == review));
@@ -235,6 +255,33 @@ public partial class GameDatabaseContext // Relations
         
         return review;
     }
+
+    public bool ReviewRatingExistsByUser(GameUser user, GameReview review)
+    {
+        RateReviewRelation? rel =
+            this.RateReviewRelations.FirstOrDefault(relation => relation.Review == review && relation.User == user);
+
+        if (rel == null)
+        {
+            return false;
+        }
+        
+        return true;
+    }
+    
+    public bool ReviewRatingExistsByUserWithSameRating(GameUser user, GameReview review, RatingType newRating)
+    {
+        RateReviewRelation? rel =
+            this.RateReviewRelations.FirstOrDefault(relation => relation.Review == review && relation.User == user && relation._ReviewRatingType == (int) newRating);
+        
+        if (rel == null)
+        {
+            return false;
+        }
+        
+        return true;
+    }
+
     
     private RateLevelRelation? GetRateRelationByUser(GameLevel level, GameUser user)
         => this.RateLevelRelations.FirstOrDefault(r => r.User == user && r.Level == level);
--- a/Refresh.GameServer/Database/GameDatabaseContext.Relations.cs
+++ b/Refresh.GameServer/Database/GameDatabaseContext.Relations.cs
@@ -196,7 +196,7 @@ public partial class GameDatabaseContext // Relations
     {
         RateReviewRelation reviewRelation = new()
         {
-            _RatingType = (int)rt,
+            RatingType = rt,
             Review = review,
         };
 
@@ -204,7 +204,6 @@ public partial class GameDatabaseContext // Relations
         {
             this.RateReviewRelations.Add(reviewRelation);
         });
-        
     }
 
     public ReviewRating GetRatingForReview(GameReview review)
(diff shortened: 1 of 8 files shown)


Refresh.GameServer/Database/GameDatabaseContext.Relations.cs
```diff
@@ -192,6 +192,81 @@ public void ClearQueue(GameUser user)
 
     #region Rating and Reviewing
 
+    public void RateReview(GameReview review, RatingType rt, GameUser user)
```
jvyden, line 195: Variable names shouldn't be shortened to this degree. `ratingType` is just fine.

Refresh.GameServer/Database/GameDatabaseContext.Relations.cs
```diff
@@ -192,6 +192,81 @@ public void ClearQueue(GameUser user)
...
+    {
+        // if rating type is neutral, remove previous review rating by user 
+        if (rt == RatingType.Neutral && this.ReviewRatingExistsByUser(user, review))
+        {
+            RateReviewRelation rel =
```
jvyden, line 200: Same thing; we use `relation` in other parts...

Refresh.GameServer/Database/GameDatabaseContext.Relations.cs
```diff
@@ -192,6 +192,81 @@ public void ClearQueue(GameUser user)
...
+                this.RateReviewRelations.First(relation => relation.Review == review && relation.User == user);
```
jvyden, line 201: ...but for LINQ we typically do first character, so `r => ...` in this case. `x => ...` is also a common pattern I see.

Refresh.GameServer/Database/GameDatabaseContext.Relations.cs
```diff
@@ -192,6 +192,81 @@ public void ClearQueue(GameUser user)
...
+            this.Write(() => this.RateReviewRelations.Remove(rel));
+
+            return;
+        }
+        // swap rating
```
jvyden, line 206:
```suggestion
        // If the relation already exists, simply set the new rating
```

Refresh.GameServer/Database/GameDatabaseContext.Relations.cs
```diff
@@ -192,6 +192,81 @@ public void ClearQueue(GameUser user)
...
+
+    public GameReview GetReviewByUsernameForLevelId(string userName, int levelId)
+    {
+        GameLevel level = this.GameLevels.First(gameLevel => gameLevel.LevelId == levelId);
+        GameUser user = this.GameUsers.First(gameUser => gameUser.Username == userName);
+        
+        GameReview review = this.GameReviews.First((gameReview => gameReview.Publisher == user && gameReview.Level == level));
+        
+        return review;
+    }
```
jvyden, line 257: In cases where you haven't verified if a review actually exists, you should use `FirstOrDefault` and handle nullable cases like this:
```csharp
    public GameReview? GetReviewByUsernameForLevelId(string userName, int levelId)
    {
        GameLevel? level = this.GameLevels.FirstOrDefault(gameLevel => gameLevel.LevelId == levelId);
        if(level == null) return null;
        GameUser? user = this.GameUsers.FirstOrDefault(gameUser => gameUser.Username == userName);
        if(user == null) return null;
        
        GameReview? review = this.GameReviews.FirstOrDefault(gameReview => gameReview.Publisher == user && gameReview.Level == level);
        
        return review;
    }
```

But I wouldn't do that in this case; I would take `GameUser` and `GameLevel` in directly; typically endpoints handle missing elements. But definitely handle the case where the end `GameReview` doesn't exist.

Refresh.GameServer/Database/GameDatabaseContext.Relations.cs
```diff
@@ -192,6 +192,81 @@ public void ClearQueue(GameUser user)
...
+
+    public bool ReviewRatingExistsByUser(GameUser user, GameReview review)
+    {
+        return this.RateReviewRelations.Any(relation => relation.Review == review && relation.User == user);
+    }
+    
+    public bool ReviewRatingExists(GameUser user, GameReview review, RatingType newRating)
+    {
+        return this.RateReviewRelations.Any(relation => relation.Review == review && relation.User == user && relation._ReviewRatingType == (int) newRating);
+    }
```
jvyden, line 267: What's the use for this?

Refresh.GameServer/Endpoints/Game/ReviewEndpoints.cs
```diff
@@ -124,4 +124,29 @@ IDateTimeProvider timeProvider
 
         return OK;
     }
+    
+    [GameEndpoint("rateReview/user/{levelId}/{userName}", HttpMethods.Post)]
+    public Response SubmitReviewRating(RequestContext request, GameDatabaseContext db, GameUser user, int levelId, string userName)
```
jvyden, line 129: For consistency:
```suggestion
    public Response SubmitReviewRating(RequestContext request, GameDatabaseContext database, GameUser user, int levelId, string username)
```

Refresh.GameServer/Types/Reviews/ReviewRating.cs
```diff
@@ -0,0 +1,7 @@
+﻿namespace Refresh.GameServer.Types.Reviews;
+
+public class ReviewRating
+{
+    public int PostiveRating { get; set; } = 0;
+    public int NegativeRating { get; set; } = 0;
+}
```
jvyden, line 7: I actually like this solution, to the point where I want to refactor a few things across the codebase to better implement this.

Could use a couple things to achieve that:
- Use a generic name, something akin to `DatabaseRatings` (similar to `DatabaseList`)
- XMLDoc to explain the usecase

jvyden (changes requested):
Looks pretty damn good. I like the use of `ReviewRating` to help with DB calls, I'll definitely have to use that in other parts of the codebase.

Aside from that I just have a few minor nitpicks relating to consistency and a couple questions:

Refresh.GameServer/Database/GameDatabaseContext.Relations.cs
jvyden: Ah, the different naming of the function and the naming of the `newRating` parameter confused me.

I think it's okay if this is an overload of `ReviewRatingExistsByUser`. But definitely change `newRating` to just `rating`.

Refresh.GameServer/Types/Reviews/ReviewRating.cs
speedyman08: does ObjectRating sound good?

speedyman08: nevermind

jvyden (changes requested):
Seems to work okay, but the problem is we're not passing back the `YourThumb` result which tells the game we rated a comment.

So you can end up rating a comment "twice" after a game restart/cache expiry which can lead to confusion:
image

See `SerializedGameReview.FromOld` to see where you should set it

Refresh.GameServer/Types/Reviews/SerializedGameReview.cs
```diff
@@ -48,16 +49,30 @@ public class SerializedGameReview : IDataConvertableFrom<SerializedGameReview, G
...
+        // if it is rated by current user
+        // then get that rating and set YourThumb to that
+
+        if (isRated) {
+            // the review rating already exists hence the ! ignores
+            RateReviewRelation? reviewRating =
+                dataContext.Database.GetRateReviewRelationForReview(dataContext.User!, review);
+
+            userRatingType = reviewRating!.RatingType;
+        }
```
jvyden, line 74: This can be massively simplified to something like this:
```suggestion
        RatingType userRatingType = dataContext.Database.GetRateReviewRelationForReview(dataContext.User!, review)?.RatingType ?? RatingType.Neutral;
```

jvyden (changes requested):

jvyden (approved):

PR #575, opened 2024-07-23
Worker + Discord Webhook refactor

### LINKED ISSUE #574: Refactor `DiscordIntegrationWorker` to use API data types internally

jvyden:
- Refactor workers to use DataContext
    This makes Workers compatible with `IDataConvertable`. Should help us expand Workers in the future as a side bonus. Admittedly, this causes a bit of jank particularly in tests and having to store the match service, but I think it should be okay.
- Convert to API data types in DiscordIntegrationWorker
    This allows us to get the same results as the API (e.g. applying PSP path, cropped asset URLs, etc.) for free, avoiding code duplication. Closes #574.


Beyley (approved):
pona

Issue #260, opened 2023-11-08
Tags & Labels


jvyden:
https://docs.littlebigrefresh.com/labels.html

Arcadius2006:
when is this coming mr. jvyman

PR #585, opened 2024-07-26
Implement level tagging

Beyley:
This is the first part of #260, implementing in-game level tagging for LBP1.

This also exposes an APIv3 endpoint for getting a list of levels by tag, and also shows the tags of a level on API level responses.

### DIFF
--- a/Refresh.GameServer/Database/GameDatabaseProvider.cs
+++ b/Refresh.GameServer/Database/GameDatabaseProvider.cs
@@ -70,6 +70,7 @@ public class GameDatabaseProvider : RealmDatabaseProvider<GameDatabaseContext>
         typeof(GameReview),
         typeof(DisallowedUser),
         typeof(RateReviewRelation),
+        typeof(TagLevelRelation),
     };
 
     public override void Warmup()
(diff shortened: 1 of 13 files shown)


Refresh.GameServer/Database/GameDatabaseProvider.cs
```diff
@@ -70,6 +70,7 @@ protected GameDatabaseProvider(IDateTimeProvider time)
         typeof(GameReview),
         typeof(DisallowedUser),
         typeof(RateReviewRelation),
+        typeof(TagLevelRelation),
```
jvyden, line 73: no db schema bump?

Refresh.GameServer/Types/Levels/Tags.cs
```diff
@@ -0,0 +1,193 @@
+using System.Text;
+
+namespace Refresh.GameServer.Types.Levels;
+
+public enum Tags
```
jvyden, line 5: This needs to be renamed to `Tags`; enums should never be plural. Especially when they're not flags.

Could also be a `byte`.

jvyden (changes requested):

Refresh.GameServer/Database/GameDatabaseProvider.cs
Beyley: realm doesn't need it for just adding new types, it only matters when an existing type changes or gets removed(?)

Refresh.GameServer/Types/Levels/Tags.cs
Beyley: i couldnt point a finger on why the name felt so wrong to say/type, will change to `Tag`, and make it a `byte`

jvyden (approved):

PR #590, opened 2024-07-28
CoolLevelsWorker: Use an exponential decay function for CR

Beyley:
This causes cool levels to *much* more aggressively favor levels from the past month or two, and past that the decay quickly converges with 0. Removing the bottom limit makes sure that cool rating will always go down, even if an older level gets popular again.

Cool Levels after this PR:

image


jvyden (approved):
Damn, that looks good now.

PR #598, opened 2024-07-30
Add support for adventures

jvyden:
Pretty straight-forward. Includes a category that only shows adventures levels as a side-bonus.

This also checks against the `GameAsset` type to see if the RootLevel is the expected type, returning an error if so. This only happens if the `GameAsset` is available; the only case where this would happen would be when re-uploading an adventure/level from the dry archive.

### DIFF
--- a/Refresh.GameServer/Endpoints/Game/Levels/PublishEndpoints.cs
+++ b/Refresh.GameServer/Endpoints/Game/Levels/PublishEndpoints.cs
@@ -1,14 +1,17 @@
 using Bunkum.Core;
 using Bunkum.Core.Endpoints;
+using Bunkum.Core.Endpoints.Debugging;
 using Bunkum.Core.Responses;
 using Bunkum.Listener.Protocol;
 using Bunkum.Protocols.Http;
 using Refresh.Common.Constants;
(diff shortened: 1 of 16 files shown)


Refresh.GameServer/Endpoints/Game/Levels/PublishEndpoints.cs
```diff
@@ -62,13 +66,28 @@ private static bool VerifyLevel(GameLevelRequest body, DataContext dataContext)
...
+        {
+            context.Logger.LogInfo(RefreshContext.Publishing, "Failed to verify root level");
+            return null;
+        }
+
+        if (body.Slots != null)
+        {
+            foreach (GameLevelRequest innerLevel in body.Slots)
+            {
+                if (VerifyLevel(body, dataContext)) continue;
```
Beyley, line 79: Did you mean to do this?
```suggestion
            foreach (GameLevelRequest innerLevel in body.Slots)
            {
                if (VerifyLevel(innerLevel, dataContext)) continue;
```

Beyley (changes requested):

Beyley (approved):

PR #606, opened 2024-08-03
Exclude level author's relations from cool levels

jvyden:
More anti-sackattack1 and pro-crankylandlord propaganda.

In English, this means that people who heart their own levels won't be able to gain CR from doing so.


Beyley (approved):
wait did that person actually do that to try to boost their spam onto cool levels? wow

jvyden:
I think so... at least it seems likely. It's probably more worth it to fix the flaw in the first place than to punish or anything for it.

PR #610, opened 2024-08-04
Migrate all unix timestamps to DateTimeOffsets

jvyden:
This probably breaks ancient Refresh database migrations (I tried to make sense of any of this shit but honestly couldn't.) but I don't think we care about that. We really need Postgres.

Might also be worth considering dropping old migrations and moving to a new system for them in the meantime.


Refresh.GameServer/Database/Activity/GameDatabaseContext.ActivityWrite.cs
```diff
@@ -18,7 +18,7 @@ public Event CreateLevelUploadEvent(GameUser userFrom, GameLevel level)
         {
             EventType = EventType.LevelUpload,
             StoredDataType = EventDataType.Level,
-            Timestamp = this._time.TimestampMilliseconds,
+            Timestamp = this._time.Now,
```
Beyley, line 21: This isnt a UTC timestamp, so when its stored in realm it will drop the offset. `SystemDateTimeProvider` should be modified to return a UTC timestamp for `.Now`, see [here](https://github.com/LittleBigRefresh/Refresh/blob/3b6d939baae5700aee3eb636510957eb2a3a27dc/Refresh.GameServer/Time/SystemDateTimeProvider.cs#L10)

Beyley (changes requested):
aside from that one thing, this looks good

Refresh.GameServer/Database/Activity/GameDatabaseContext.ActivityWrite.cs
jvyden: ...This has broken this entire time?

Beyley: seemingly so

Beyley (approved):

PR #623, opened 2024-08-07
Allow conversion of assets not in the database to images

jvyden:
Implemented as discussed on [Discord](https://discord.com/channels/1049223665243389953/1049225857350254632/1270832582501597378).

For ease of testing, use the included new `IDataStore` named `RemoteRefreshDataStore` to use prod as a storage backend, then look at reuploads from the archive. Notice that they now have icons.


Beyley (approved):
all works locally

PR #630, opened 2024-08-11
Show levels from the last active contest after a contest has ended

jvyden:
Right now, our contest category is empty, but people still care about #LBWSBR. This makes it so that the contest category will *always* have levels from that contest, despite the fact it's over.

### DIFF
--- a/Refresh.GameServer/Database/GameDatabaseContext.Contests.cs
+++ b/Refresh.GameServer/Database/GameDatabaseContext.Contests.cs
@@ -52,6 +52,15 @@ public partial class GameDatabaseContext // Contests
             .FirstOrDefault();
     }
     
+    public GameContest? GetLastActiveContest()
+    {
+        DateTimeOffset now = this._time.Now;
+        return this.GameContests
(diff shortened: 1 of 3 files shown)


Refresh.GameServer/Database/GameDatabaseContext.Contests.cs
```diff
@@ -52,6 +52,15 @@ public IEnumerable<GameContest> GetAllContests()
...
     }
     
+    public GameContest? GetLastActiveContest()
+    {
+        DateTimeOffset now = this._time.Now;
+        return this.GameContests
+            .Where(c => c.EndDate <= now)
+            .OrderByDescending(c => c.EndDate)
+            .FirstOrDefault();
+    }
```
Beyley, line 62: When i read the name "get last active contest" I imagine that it would get the contest which was most recently active (including currently running contests), i think `GetLatestCompletedContest` is a much more clear name

Beyley (changes requested):

PR #638, opened 2024-08-13
Flag modded levels + per-user mod filtering

Beyley:
Adds a flag to all levels to state whether they contain modded content of any kind. Also adds a user setting + command to hide all modded content from the game, if the user does not wish to see any modded content in level results.

This PR also fixes MATT asset uploads and optimizes our realm migrations.

### DIFF
--- a/Refresh.GameServer/Database/GameDatabaseProvider.cs
+++ b/Refresh.GameServer/Database/GameDatabaseProvider.cs
@@ -33,7 +33,7 @@ public class GameDatabaseProvider : RealmDatabaseProvider<GameDatabaseContext>
         this._time = time;
     }
 
-    protected override ulong SchemaVersion => 143;
+    protected override ulong SchemaVersion => 145;
 
     protected override string Filename => "refreshGameServer.realm";
     
@@ -97,8 +97,8 @@ public class GameDatabaseProvider : RealmDatabaseProvider<GameDatabaseContext>
         IQueryable<dynamic>? oldUsers = migration.OldRealm.DynamicApi.All("GameUser");
         IQueryable<GameUser>? newUsers = migration.NewRealm.All<GameUser>();
 
-        // only run all this if old version < 134, since thats the last time these were changed (update this accordingly)
-        if (oldVersion < 134)
+        // only run all this if old version < 145, since thats the last time these were changed (update this accordingly)
+        if (oldVersion < 145)
             for (int i = 0; i < newUsers.Count(); i++)
             {
                 dynamic oldUser = oldUsers.ElementAt(i);
@@ -216,6 +216,9 @@ public class GameDatabaseProvider : RealmDatabaseProvider<GameDatabaseContext>
                         });
                     }
                 }
+
+                if (oldVersion < 145)
+                    newUser.ShowModdedContent = true;
             }
 
         IQueryable<dynamic>? oldLevels = migration.OldRealm.DynamicApi.All("GameLevel");
--- a/Refresh.GameServer/Database/GameDatabaseProvider.cs
+++ b/Refresh.GameServer/Database/GameDatabaseProvider.cs
@@ -33,7 +33,7 @@ public class GameDatabaseProvider : RealmDatabaseProvider<GameDatabaseContext>
         this._time = time;
     }
 
-    protected override ulong SchemaVersion => 141;
+    protected override ulong SchemaVersion => 143;
 
     protected override string Filename => "refreshGameServer.realm";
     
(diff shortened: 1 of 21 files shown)


jvyden (changes requested):
Not gonna bother reviewing... Realm migration changes should be a different PR entirely.

And also, since we found levels flagged as modded despite being made of vanilla assets (e.g. the pod), this is prone to mis-flagging and I'm not sure we should be exposing the option to hide modded levels until we figure out a solution to that.

Beyley:
Marking as draft until I split the realm changes out to another PR

Beyley:
Misflagging of vanilla assets has been solved. This PR also relaxes the mod filter slightly, as it has been discovered that a couple extra asset types can (very rarely) be uploaded by the game on an unmodified copy. After those changes, only ~10% of all levels on prod are marked as "modded", compared with the ~12% before those changes.

Refresh.GameServer/Importing/AssetImporter.cs
```diff
@@ -19,80 +21,81 @@ public AssetImporter(Logger? logger = null, IDateTimeProvider? timeProvider = nu
...
 
-            assets.Add(newAsset);
-            this.Info($"Processed {newAsset.AssetType} asset {hash} ({AssetSafetyLevelExtensions.FromAssetType(newAsset.AssetType, newAsset.AssetFormat)})");
+                tasksDone[threadId] = true;
+            }).Start();
         }
-        
-        database.AddOrUpdateAssetsInDatabase(assets);
 
+        SpinWait.SpinUntil(() => tasksDone.All(done => done));
```
jvyden, line 84: Wouldn't this hang if a thread throws internally and exits? Also disagree with spinning on a long-running operation.

It's better if we use tasks here. Store `Task[]` instead of `bool[]`, then use [`Task.WaitAll`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.waitall?view=net-8.0#system-threading-tasks-task-waitall(system-threading-tasks-task())), which also has the bonus of handling `Task`s that throw by bubbling up an `AggregateException`. It's true that we don't disturb the thread pool by using dedicated threads, but we can also just hint to the pool that the task will be long-running and it'll be handled much better.

Refresh.Common/Helpers/HexHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+    
+    /// <summary>
+    /// Converts a hex string to a byte array
+    /// </summary>
+    /// <param name="hex"></param>
+    /// <returns></returns>
+    /// <exception cref="Exception"></exception>
+    [MethodImpl(MethodImplOptions.AggressiveOptimization)]
+    public static byte[] HexStringToBytes(string hex) {
+        if (hex.Length % 2 != 0)
```
jvyden, line 34: Should run formatting over this

Refresh.Common/Helpers/HexHelper.cs
```diff
@@ -0,0 +1,71 @@
...
+            throw new Exception("The hex string must have an even number of bytes");
```
jvyden, line 35:
```suggestion
            throw new FormatException("The hex string is invalid as it contains an odd number of bytes.");
```

Refresh.GameServer/Services/CommandService.cs
```diff
@@ -115,6 +115,16 @@ public void HandleCommand(CommandInvocation command, GameDatabaseContext databas
...
+            case "showmoddedcontent":
+            {
+                database.SetShowModdedContent(user, true);
+                break;
+            }
+            case "hidemoddedcontent":
+            {
+                database.SetShowModdedContent(user, false);
+                break;
+            }
```
jvyden, line 127: It's important that we use as few characters as possible; typing on console is a pain in the ass.
```suggestion
            case "showmods":
            {
                database.SetShowModdedContent(user, true);
                break;
            }
            case "hidemods":
            {
                database.SetShowModdedContent(user, false);
                break;
            }
```

Refresh.GameServer/Services/VanillaHashChecker.cs
```diff
@@ -0,0 +1,34 @@
...
+        }
+    }
+    
+    public static bool IsVanillaHash(string hash)
+    {
+        byte[] hashArr = HexHelper.HexStringToBytes(hash);
+
+        return VanillaHashes.Contains(hashArr);
+    }
+}
```
jvyden, line 34: I would move this to Refresh.Common, tbh.

Refresh.GameServer/Services/VanillaHashChecker.cs
```diff
@@ -0,0 +1,34 @@
...
+        using Stream stream = ResourceHelper.StreamFromResource("Refresh.GameServer.Resources.all_hashes.txt");
+        
+        StreamReader reader = new(stream);
+
+        //Read all the lines
+        while (reader.ReadLine() is {} line)
+        {
+            //Parse out and add each line to the set
+            VanillaHashes.Add(HexHelper.HexStringToBytes(line));
+        }
```
jvyden, line 25: This is a micro-optimization for sure, but this makes me wonder if it's a good idea to store the hashes in binary, and then just read the bytes directly into `VanillaHashes` so we can skip this conversion.

I guess the drawback is that the file wouldn't be human readable? I'm not sure if that's really that bad though.

Refresh.GameServer/Types/Levels/GameLevel.cs
```diff
@@ -48,6 +48,11 @@ public partial class GameLevel : IRealmObject, ISequentialId
     public bool TeamPicked => this.DateTeamPicked != null;
     public DateTimeOffset? DateTeamPicked { get; set; }
     
+    /// <summary>
+    /// Whether any asset in the dependency tree is considered "modded"
+    /// </summary>
+    public bool Modded { get; set; }
```
jvyden, line 54: Same here, should use `IsModded`

Refresh.GameServer/Extensions/LevelEnumerableExtensions.cs
```diff
@@ -38,6 +38,9 @@ public static IQueryable<GameLevel> FilterByLevelFilterSettings(this IQueryable<
         if (levelFilterSettings.ExcludeMyLevels && user != null)
             levels = levels.Where(l => l.Publisher != user);
 
+        if (user is { ShowModdedContent: false })
+            levels = levels.Where(l => !l.Modded);
+
```
jvyden, line 43: We'll probably want to also have an option in `levelFilterSettings` for a temporary override when browsing on the website.

jvyden (changes requested):

Refresh.GameServer/Services/VanillaHashChecker.cs
Beyley: I switched it to a `FrozenSet`, thanks for the reminder of its existance. I think if this does end up being a problem in the future, we can switch to a binary based format. For reference, the generation code for the file is present [here](https://github.com/Beyley/zishzingers/commit/c3a99ab7c084ceb34a67d9890c778f20d1cb626b)

Refresh.GameServer/Database/GameDatabaseProvider.cs
```diff
@@ -333,6 +336,42 @@ protected override void Migrate(Migration migration, ulong oldVersion)
...
+                            bool modded = false;
+                            TraverseDependenciesRecursively(asset, (hash, gameAsset) =>
+                            {
+                                if (gameAsset != null && (gameAsset.AssetFlags & AssetFlags.Modded) != 0)
+                                    modded = true;
+                            });
+                            newLevel.Modded = modded;
+                        }
+                    }
+                }
```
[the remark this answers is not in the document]
Beyley: I'm not exactly sure how the implementation of that would go (another Config to store the version maybe?), but given i did add the `--flag-modded-levels` option to run it manually, I think for the purposes of not making this PR too big, it should be fine to leave it out for now and tackle automated asset re-imports and such in a different PR.

jvyden (approved):
I like the state of where this is, so I'm gonna merge. Hopefully false-flags don't cause problems.

Issue #565, opened 2024-07-19
Audit Log


Beyley:
Refresh should store an audit log of all moderation actions, so bans, level deletions, metadata changes by admins/curators, etc. into a viewable log accessible through the API. There should also be discord webhook integration so that these actions can be automatically logged to a discord channel aswell.

This is useful for admins so that there is a consolidated place to keep track of past actions taken against a particular user, but could also be useful for server owners, who may want to make this viewable through a public discord channel, to make the gameserver moderation more transparent.

Toastbrot236:
Things left to do before this can be closed:
- Actually log moderation actions
- Ability for moderators to see all moderation actions, and other users to be able to (only) see actions taken against them
- Config option which would allow logging actions to a discord channel

PR #644, opened 2024-08-24
Introduce integrations with AIPI for automated content scanning

jvyden:
Rough port from DO code to integrate AI image scanning into Refresh. This should help reduce the photo endpoint abuse we've been seeing from DT, while using the same existing infrastructure that works incredibly well.

This does not scan every uploaded texture as this would increase server load significantly. It will only scan under certain circumstances:
- for uploaded photos
- for images uploaded via the API

This also brings in the option to automatically restrict users who upload images with banned tags if configured. This will be left off on production while I gauge if this is going to false-flag for normal non-abuse activity.

Oh, this also introduces support for a staff-only webhook. This could be used for a non-persisted audit log if we wanted to use that as a solution for #565.

### DIFF
--- a/Refresh.GameServer/Services/AipiService.cs
+++ b/Refresh.GameServer/Services/AipiService.cs
@@ -1,0 +1,161 @@
+using System.Diagnostics;
+using System.Net.Http.Json;
+using Bunkum.Core.Services;
+using JetBrains.Annotations;
+using NotEnoughLogs;
+using Refresh.GameServer.Configuration;
+using Refresh.GameServer.Importing;
+using Refresh.GameServer.Types.Assets;
+using Refresh.GameServer.Types.Data;
+using SixLabors.ImageSharp;
+using SixLabors.ImageSharp.Formats;
+using SixLabors.ImageSharp.Processing;
+
+namespace Refresh.GameServer.Services;
+
+// Referenced from DO.
+public class AipiService : EndpointService
+{
+    private readonly HttpClient _client;
+    private readonly IntegrationConfig _config;
+    private readonly DiscordStaffService? _discord;
+
+    private readonly ImageImporter _importer;
+    
+    [UsedImplicitly]
+    public AipiService(Logger logger, IntegrationConfig config, ImportService import, DiscordStaffService discord) : base(logger)
+    {
+        this._discord = discord;
+        this._config = config;
+
+        this._client = new HttpClient
+        {
+            BaseAddress = new Uri(config.AipiBaseUrl),
+        };
+
+        this._importer = import.ImageImporter;
+    }
+
+    public override void Initialize()
+    {
+        if (!this._config.DiscordStaffWebhookEnabled)
+        {
+            this.Logger.LogWarning(RefreshContext.Aipi, 
+                "The Discord staff webhook is not enabled, but AIPI is. This is probably behavior you don't want.");
+        }
+        this.TestConnectivityAsync().Wait();
+    }
+
+    private async Task TestConnectivityAsync()
+    {
+        try
+        {
+            HttpResponseMessage response = await this._client.GetAsync("/");
+            string content = await response.Content.ReadAsStringAsync();
+
+            if (response.IsSuccessStatusCode && content == "AIPI scanning service")
+                this.Logger.LogInfo(RefreshContext.Aipi, "AIPI appears to be working correctly");
+            else
+                this.Logger.LogError(RefreshContext.Aipi,
+                    $"AIPI seems to be down. Status code: {response.StatusCode}, content: {content}");
+        }
+        catch (Exception e)
+        {
+            this.Logger.LogError(RefreshContext.Aipi, "AIPI connection failed: {0}", e.ToString());
+        }
+    }
+    
+    private async Task<TData> PostAsync<TData>(string endpoint, Stream data)
+    {
+        HttpResponseMessage response = await this._client.PostAsync(endpoint, new StreamContent(data));
+        AipiResponse<TData>? aipiResponse = await response.Content.ReadFromJsonAsync<AipiResponse<TData>>();
+        
+        if (aipiResponse == null) throw new Exception("No response was received from the server.");
+        if (!aipiResponse.Success) throw new Exception($"{response.StatusCode}: {aipiResponse.Reason}");
+
+        return aipiResponse.Data!;
+    }
+    
+    private async Task<Dictionary<string, float>> PredictEvaAsync(Stream data)
+    {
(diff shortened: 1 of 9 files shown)


Refresh.GameServer/Services/AipiService.cs
```diff
@@ -0,0 +1,161 @@
...
+        }
+    }
+    
+    private async Task<TData> PostAsync<TData>(string endpoint, Stream data)
+    {
+        HttpResponseMessage response = await this._client.PostAsync(endpoint, new StreamContent(data));
+        AipiResponse<TData>? aipiResponse = await response.Content.ReadFromJsonAsync<AipiResponse<TData>>();
+        
+        if (aipiResponse == null) throw new Exception("No response was received from the server.");
+        if (!aipiResponse.Success) throw new Exception($"{response.StatusCode}: {aipiResponse.Reason}");
```
Beyley, line 74: There should be an `AipiException` class, or the exception text should elaborate that it comes from AIPI, it makes grepping logs much easier.

Refresh.GameServer/Services/AipiService.cs
```diff
@@ -0,0 +1,161 @@
...
+            return false;
+
+        // import the asset as png
+        bool isPspAsset = asset.AssetHash.StartsWith("psp/");
+        string hash = isPspAsset ? "psp/" + asset.AssetHash : asset.AssetHash;
+
+        if (!context.DataStore.ExistsInStore("png/" + asset.AssetHash))
+        {
+            this._importer.ImportAsset(asset.AssetHash, isPspAsset, asset.AssetType, context.DataStore);
+        }
```
Beyley, line 134: That `hash` variable is unused, and I believe can be removed, given `ImportAsset` will handle PSP assets for us.

I'm also really starting to dislike that `psp` is something we have to prepend everywhere, it really feels like a hack. It's out of scope for this PR, but at some point I'm going to find a modular solution to the issue of "game locked assets".

Refresh.GameServer/Services/AipiService.cs
```diff
@@ -0,0 +1,161 @@
...
+        // guard the fact that assets have an owner
+        Debug.Assert(asset.OriginalUploader != null, $"Asset {asset.AssetHash} had no original uploader when trying to scan");
+        if (asset.OriginalUploader == null)
+            return false;
+
+        // import the asset as png
+        bool isPspAsset = asset.AssetHash.StartsWith("psp/");
+        string hash = isPspAsset ? "psp/" + asset.AssetHash : asset.AssetHash;
+
+        if (!context.DataStore.ExistsInStore("png/" + asset.AssetHash))
```
Beyley, line 131: We should do this `"png/" + AssetHash` once and store the result, since we need this string later down aswell, and right now we allocate both times.

Refresh.GameServer/Services/DiscordStaffService.cs
```diff
@@ -0,0 +1,75 @@
...
+        
+        if(config.DiscordStaffWebhookEnabled)
+            this._client = new DiscordWebhookClient(config.DiscordStaffWebhookUrl);
+    }
+    
+    private string GetAssetUrl(string hash)
+    {
+        return $"{this._externalUrl}/api/v3/assets/{hash}/image";
+    }
+
```
Beyley, line 37: This code is duplicated from [here](https://github.com/LittleBigRefresh/Refresh/blob/6406dfb24effb8f87d5f1d1f0e6950a536f622af/Refresh.GameServer/Workers/DiscordIntegrationWorker.cs#L39), maybe we should have a helper method?

Refresh.GameServer/Services/DiscordStaffService.cs
```diff
@@ -0,0 +1,75 @@
...
+    
+    private string GetAssetUrl(string hash)
+    {
+        return $"{this._externalUrl}/api/v3/assets/{hash}/image";
+    }
+    
+    private string GetAssetInfoUrl(string hash)
+    {
+        return $"{this._externalUrl}/api/v3/assets/{hash}/image";
+    }
```
Beyley, line 41: Did you mean this? The function is labelled as `AssetInfoUrl`.
```suggestion
    private string GetAssetInfoUrl(string hash)
    {
        return $"{this._externalUrl}/api/v3/assets/{hash}";
    }
```

Refresh.GameServer/Services/DiscordStaffService.cs
```diff
@@ -0,0 +1,75 @@
...
+            .WithTitle($"AI Analysis of `{asset.AssetHash}`");
+        
+        foreach ((string tag, float confidence) in results.OrderByDescending(r => r.Value).Take(25))
+        {
+            string tagFormatted = this._config.AipiBannedTags.Contains(tag) ? $"{tag} (flagged!)" : tag;
+            string confidenceFormatted = confidence.ToString("0.00%");
+            builder.AddField(tagFormatted, confidenceFormatted, true);
+        }
+        
+        this.PostMessage($"Prediction result for {author.Username}:", [builder.Build()]);
```
Beyley, line 73: This message should be a bit clearer
```suggestion
        this.PostMessage($"Prediction result for {asset.AssetHash} ({author.Username}):", [builder.Build()]);
```

Beyley (changes requested):
Functionality looks good, aside from what seems to be a typo with one of the API links. 
Not all of these comments necessarily need to be resolved before merge, if you would like to get this out sooner.

Refresh.GameServer/Services/DiscordStaffService.cs
jvyden: It's more of a formality for logging; all the content you actually want is in the embed. But I guess that's fine.

Refresh.GameServer/Services/AipiService.cs
jvyden: We are talking bytes here... No thanks

Refresh.GameServer/Services/DiscordStaffService.cs
jvyden: I'm probably going to move webhook logic to DataContext via service in a later PR. Fine for now.

Refresh.GameServer/Services/AipiService.cs
jvyden: Didn't catch that.

Refresh.GameServer/Services/AipiService.cs
jvyden: Errors are more of a formality; you're unlikely to actually get one (as it's mostly going to be a server error which is logged in more detail anyways), so this is fine for now.

PR #649, opened 2024-08-31
LBP1 Playlists

Beyley:
You can only create playlists in-game when latest retail LBP1 has the memory address `0x0074c2c0` patched to be `0x01` instead of `0x00`. In the EBOOT this is located at address `0x0073c2c0`. Updates to Refresher and the OFW patchers will need to be made to enable this feature. For testing you also need to make sure that the `/lbp/network_settings.nws` endpoint is being accessed correctly, since playlists require the `AlexDB` option to be set to function.

Luckily the game is perfectly happy viewing and playing other's playlists even if you don't have the feature enabled yourself. So there's no worry about un-patched users having issues.

I would like to tackle APIv3 support for playlists in a future PR, but aside from that, this is a complete implementation of the feature for LBP1.

It should be possible to represent this (at least close enough) in LBP3 by just not showing sub-playlists, and showing all playlists by a user, instead of just playlists inside of your root playlist. For LBP2/PSP/Vita we can just display playlists as a slot results override you trigger from the API, or as a special cased `playlist:name` in text search, although we won't be able to display sub-playlists there either.

As a basic technical summary of how this is layed out, all players contain a "Root Playlist". In LBP1, when you view your "My Playlists" view, it will create a root playlist for you, if you don't have one. All playlists you create in the future are then considered "sub-playlists" of that root playlist, or of some other playlist.

The game allows you to add any playlist (eg. other's playlists aswell) as a sub-playlist to any of your playlists (except your root playlist). And it also allows you to add any `user` slot to any of your playlists (except your root playlist), as a sub-level of the playlist.

The original developer's intention for viewing a user's playlists is that there is a button on the profile labeled "Playlists", and it would show the contents of the root playlist, however I found this to be rather uncreative, so in LBP1 I've injected the contents of your root playlist into your user level list as "sticky" slot results, so it always appears regardless of the page. This seems to work fairly well, and doesn't impact pagination from what I can see.

PR #652, opened 2024-09-03
Add presence server

Beyley:
This server is responsible for one thing only, which is telling the client to load the slot listing for any user slot you send it. Retail LBP2 and LBP3 both implement this, but both require [special care](https://discord.com/channels/1049223665243389953/1049225857350254632/1280019755482218588) to make it use a separate domain (which is important if you want to protect your main gameserver behind cloudflare!)

This is a fairly basic implementation of the server, but should be sufficient for even large amounts of connected players.

Marking as draft until #649 is merged and hashed live play now is added, since this branch is based off of that one (didnt wanna deal with merge conflicts on this PR).

Thanks to aidan for figuring all this out

Example of server in action:

https://github.com/user-attachments/assets/d790a570-c291-4005-b3cf-daf6a95a9f51


Beyley:
I added a field to `ApiExtendedGameUserResponse` for whether the user is on the presence server, this should allow frontends to switch the terminology of popups/buttons to reflect the differing behaviour.

Refresh.PresenceServer/Server/PresenceServer.cs
```diff
@@ -0,0 +1,303 @@
...
+                    tcpClient.SendBufferSize = 256;
+                    tcpClient.ReceiveBufferSize = 256;
+
+                    GameClient gameClient = new(tcpClient);
+                    this._clients.Add(gameClient);
+                    this._logger.LogInfo(PresenceCategory.Connections, "Client {0} connected.", gameClient.IpAddress);
+                }
+                catch(Exception ex)
+                {
+                    this._logger.LogWarning(PresenceCategory.Connections, "Failed to accept client. Reason {0}", ex);
```
jvyden, line 111:
```suggestion
                    this._logger.LogError(PresenceCategory.Connections, "Failed to accept client. Reason {0}", ex);
```

Refresh.PresenceServer/Server/PresenceServer.cs
```diff
@@ -0,0 +1,303 @@
...
+                            ResourceHelper.XxteaEncrypt(sendData[8..], this._key);
+                            
+                            client.TcpClient.Client.Send(sendData);
+                            
+                            this._logger.LogInfo(PresenceCategory.Connections, "Sending slot ID {0} to user {1}", slotId, client.IpAddress);
+                        }
+                        catch(Exception ex)
+                        {
+                            this._logger.LogWarning(PresenceCategory.Connections,
+                                "Failed to send packet data to {0}, reason {1}", client.IpAddress, ex);
```
jvyden, line 163:
```suggestion
                                "Failed to send packet data to {0}: {1}", client.IpAddress, ex);
```

Refresh.PresenceServer/Server/PresenceServer.cs
```diff
@@ -0,0 +1,303 @@
...
+                    break;
+                }
+                // keepalive packet
+                case 0x0d when read[1] == 0x0a && read.Length == 2:
+                    this._logger.LogDebug(PresenceCategory.Connections, "Keepalive from {0}", gameClient.IpAddress);
+
+                    break;
+                default:
+                    this._logger.LogWarning(PresenceCategory.Connections,
+                        "Unknown packet from {0}, treating as basic keepalive", gameClient.IpAddress);
```
jvyden, line 257: I feel like we should log the packet in debug at least? Probably fine though

jvyden (changes requested):

Refresh.GameServer/Endpoints/Game/Handshake/AuthenticationEndpoints.cs
```diff
@@ -278,6 +281,8 @@ public Response RevokeThisToken(RequestContext context, GameDatabaseContext data
         
         // Revoke the token
         database.RevokeToken(token);
+        // Mark them as disconnected from the presence server
+        database.SetUserPresenceAuthToken(user, null);
```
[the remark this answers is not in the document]
Beyley: Not expected, but I don't think its particularly bad if that happens, since it gets cleared on game login, and gets cleared when the presence server detects a disconnection. The data of "connected to presence server" is only useful if we know the user is in-game in a lobby (that's how list play now works on refresh-web currently)
Clearing it here is just a "good samaritan" thing to do, rather than required for any use/functionality

Refresh.GameServer/Endpoints/Internal/Presence/PresenceEndpoints.cs
```diff
@@ -0,0 +1,49 @@
...
+        if (user == null)
+            return NotFound;
+        
+        database.SetUserPresenceAuthToken(user, null);
+
+        context.Logger.LogInfo(RefreshContext.Presence, $"User {user} disconnected from the presence server");
+
+        return OK;
+    }
+}
```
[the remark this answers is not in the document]
Beyley: agree, interserver communication isnt required in this way once we have postgres

Refresh.PresenceServer/Server/PresenceServer.cs
Beyley: Given the packets logged are the only ones the client is ever capable of sending, other packet data is an indication that something has gone quite wrong (eg. HTTP requests are somehow making their way through to it), so I'd prefer to keep it a warning, we can move to debug only if it starts happening frequently though

Refresh.PresenceServer/ApiServer/Authentication/SecretAuthenticationProvider.cs
```diff
@@ -0,0 +1,21 @@
...
+    public SecretAuthenticationProvider(PresenceServerConfig config)
+    {
+        this._config = config;
+    }
+
+    public Token? AuthenticateToken(ListenerContext request, Lazy<IDatabaseContext> database)
+    {
+        return request.RequestHeaders["Authorization"] == this._config.SharedSecret ? new Token() : null;
+    }
+}
```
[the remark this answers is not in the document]
Beyley: Moved to an endpoint

Refresh.Common/Constants/EndpointRoutes.cs
```diff
@@ -0,0 +1,6 @@
+namespace Refresh.Common.Constants;
+
+public static class EndpointRoutes
+{
+    public const string PresenceBaseRoute = "/_internal/presence/";
+}
```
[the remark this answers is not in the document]
Beyley: Not strictly necessary, but given this string is needed in both Refresh.GameServer and Refresh.PresenceServer, I'd rather not duplicate the string if possible, so I put it in a constant

Refresh.PresenceServer/ApiServer/Middlewares/SecretMiddleware.cs
```diff
@@ -0,0 +1,27 @@
+using Bunkum.Core.Database;
+using Bunkum.Core.Endpoints.Middlewares;
+using Bunkum.Listener.Request;
+using Refresh.PresenceServer.Server.Config;
+
+namespace Refresh.PresenceServer.ApiServer.Middlewares;
+
+public class SecretMiddleware : IMiddleware
```
jvyden, line 8: `SharedSecretAuthMiddleware`

jvyden (approved):
There's some minor syntax/naming things in here but otherwise this looks pretty good now.

PR #653, opened 2024-09-05
Send notifications when someone comments on your profile/level

### LINKED ISSUE #648: Send notifications when someone comments on your profile/level

deadly:
Suggested by @Arcadius2006 -- Closes #648


.gitignore
```diff
@@ -353,4 +353,11 @@ MigrationBackup/
...
-/data/
\ No newline at end of file
+/data/
+
+# db & settings
+
+*.realm
+*.lock
+*.json
+dataStore/
```
jvyden, line 363: If you're seeing these outside of `bin/Debug/net8.0`, you're probably running Refresh incorrectly.

We also have a `.json` file in use not relating to Realm or Bunkum/Refresh: https://github.com/LittleBigRefresh/Refresh/blob/main/global.json

Refresh.GameServer/Endpoints/Game/CommentEndpoints.cs
```diff
@@ -30,6 +30,12 @@ public Response PostProfileComment(RequestContext context, GameDatabaseContext d
         GameUser? profile = database.GetUserByUsername(username);
         if (profile == null) return NotFound;
         
+        
+        int count = database.GetNotificationCountByUser(profile);
+        if (!profile.Equals(user) && count < 10) {
```
jvyden, line 35: I like the idea behind not spamming the user with notifications, but this seems a little arbitrary...

I think the end-goal here is for users to be able to toggle which notifications they want to receive. Maybe a set of 'channels', so like:
```csharp
enum NotificationChannel : byte
{
    UserHeart,
    UserMutual,
    ScoreOvertake,
    Comment,
    // etc...
}
```

And then you'd be able to fine-tune which of those you want to receive on the website.

I'm okay with just always sending a notification for now but this is something that should be addressed in a follow-up PR either by one of us, or you if you'd like.

Refresh.GameServer/Endpoints/Game/CommentEndpoints.cs
```diff
@@ -82,6 +88,12 @@ public Response PostLevelComment(RequestContext context, GameDatabaseContext dat
         GameLevel? level = database.GetLevelByIdAndType(slotType, id);
         if (level == null) return NotFound;
 
+        
+        int count = database.GetNotificationCountByUser(level.Publisher);
+        if (!level.Publisher.Equals(user) && count < 10) {
```
jvyden, line 93: Also, we use [BSD braces](https://en.wikipedia.org/wiki/Indentation_style#Allman_style)

Refresh.GameServer/Endpoints/Game/CommentEndpoints.cs
```diff
@@ -82,6 +88,12 @@ public Response PostLevelComment(RequestContext context, GameDatabaseContext dat
...
+            database.AddNotification("New level comment", $"{user.Username} left a comment on your level: '{level.Title}!'", level.Publisher);
```
jvyden, line 94:
```suggestion
            database.AddNotification("New comment", $"{user.Username} left a comment on your level: '{level.Title}!'", level.Publisher);
```

jvyden (changes requested):

Refresh.GameServer/Endpoints/Game/CommentEndpoints.cs
Arcadius2006: HOLD ON, BY NOTIFICATIONS I MEANT RECENT ACTIVITY NOTIFICATIONS

Beyley: > HOLD ON, BY NOTIFICATIONS I MEANT RECENT ACTIVITY NOTIFICATIONS

doesn't matter for this PR, we are still going to accept it since this is a nice feature (not everyone wants to check ingame/on recent activity to see this), but you should clarify in your issues that "notifications" means "recent activity events", since those are distinctly different things.

Arcadius2006: yeah, sorry... i just didn't know what to call them really

i hope it's fine

jvyden: You can make a new issue but please stop commenting on the PR.

Refresh.GameServer/Endpoints/Game/CommentEndpoints.cs
deadly: Sorry, you're right, I shouldn't have used a magic number like that. I like the idea of toggling notifications, I'll try to work on it soon.

jvyden:
WTF did you do to the branch?

jvyden:
Try to use Github's interface for merging upstream into your branch next time... you seem to have caused some strange behavior with whatever you did to bring those commits in.. like rebased the wrong direction?

jvyden (approved):

PR #659, opened 2024-09-10
Prepare for upcoming RPCN ticket changes

jvyden:
Uses the `SignatureIdentifier` to match RPCN tickets instead of the now-deprectaed `IssuerId` of `0x33333333` to verify ticket. The old method is kept as a fallback for ease of mind purposes.

https://discord.com/channels/272035812277878785/738838647176036393/1283019875538829354

### DIFF
--- a/Refresh.GameServer/Extensions/TicketExtensions.cs
+++ b/Refresh.GameServer/Extensions/TicketExtensions.cs
@@ -1,0 +1,31 @@
+using Bunkum.Core;
+using NPTicket;
+using Refresh.GameServer.Authentication;
+
+namespace Refresh.GameServer.Extensions;
+
+public static class TicketExtensions
+{
+    public static TokenPlatform? DeterminePlatform(this Ticket ticket)
+    {
+        if (ticket.SignatureIdentifier == "RPCN" || ticket.IssuerId == 0x33333333)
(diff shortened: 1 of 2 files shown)


Refresh.GameServer/Extensions/TicketExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+using Refresh.GameServer.Authentication;
+
+namespace Refresh.GameServer.Extensions;
+
+public static class TicketExtensions
+{
+    public static TokenPlatform? DeterminePlatform(this Ticket ticket)
+    {
+        if (ticket.SignatureIdentifier == "RPCN" || ticket.IssuerId == 0x33333333)
+            return TokenPlatform.RPCS3;
```
Beyley, line 12: The message implied that there could be *more* to the signature than just "RPCN", and implies only the first four bytes of the signature will be "RPCN", are we sure `==` will hold true?

jvyden: Let me ask them.

jvyden: I'm gonna go ahead with "yes this is likely the case" for now but if that changes I'll readdress this.

PR #668, opened 2024-09-25
Change HTTPS proxy to work off redirects towards an external proxy

Beyley:
This allows us to use an external proxy, instead of the buggy proxy that I had originally written. Sadly we cant redirect directly to prod because of a bug in the original game (after a redirect it still uses the pre-redirect host header value with the new server). 

In practice, the proxy.json config will probably just point to the same machine as Refresh.HttpsProxy itself, but with a different port, which is hosting a standard nginx HTTP proxy server.

This PR also fixes the middleware order


jvyden (approved):
Sorry, this PR flew under the radar

PR #676, opened 2024-10-20
Limit verified IPs to 3

Beyley:


Refresh.GameServer/Database/RealmDbSet.cs
```diff
@@ -43,6 +43,12 @@ public void RemoveRange(IQueryable<T> objs)
     {
         this._realm.RemoveRange(objs);
     }
+    
+    public void RemoveRange(IEnumerable<T> objs)
+    {
+        foreach (T obj in objs) 
+            this._realm.Remove(obj);
+    }
```
jvyden, line 51: This is just a note but this will likely break source compatibility with EF as I don't believe this exists on a `DbSet`. Not too hard to fix when we get there.

jvyden (approved):

PR #683, opened 2024-11-02
Add API endpoint to purge all reviews by the given user

Beyley:


jvyden (approved):
In the future this should also work for individual reviews. And also for comments.

PR #697, opened 2024-12-06
Implement moving levels while decorating earth

Toastbrot236:
Implements the ability to move around levels while decorating your earth using the "Sticker & Decoration Edit Tool", which is very practical when you want to move multiple levels quickly without going through each level individually.


Toastbrot236:
No way this is already a month old.

Toastbrot236:
2 months

PR #720, opened 2025-03-06
LBP3 Playlists - attempt 2

Toastbrot236:
A recreation (and improvement to some extent) of my first lbp3 playlists fork which I deleted after merging it into my main. This PR adds playlist support for LBP3, and also fixes a way to create loops inside playlist trees in LBP1 and a related server crash. I really hope i havent forgotten anything while copying from my original branch, but I did quickly re-test this in LBP1 and 3.


Refresh.GameServer/Database/GameDatabaseContext.Playlists.cs
```diff
@@ -101,56 +170,165 @@ public void AddLevelToPlaylist(GameLevel level, GamePlaylist parent)
...
+    public int GetTotalLevelsInPlaylistCount(GamePlaylist playlist, TokenGame game) => 
+        this.LevelPlaylistRelations.Where(l => l.Playlist == playlist).AsEnumerable()
+            .Select(l => l.Level)
+            .FilterByGameVersion(game)
+            .Count();
+
+    public int GetTotalLevelsInPlaylistCount(GamePlaylist playlist) => 
+        this.LevelPlaylistRelations.Where(l => l.Playlist == playlist).AsEnumerable()
+            .Select(l => l.Level)
+            .Count();
```
jvyden, line 253: Since the level itself is unused, this might be able to be shortened to this:

```suggestion
    public int GetTotalLevelsInPlaylistCount(GamePlaylist playlist) => 
        this.LevelPlaylistRelations.Count(l => l.Playlist == playlist)
```

Refresh.GameServer/Endpoints/Game/Playlists/Lbp3PlaylistEndpoints.cs
```diff
@@ -0,0 +1,224 @@
...
+
+        int index = 0;
+        foreach (int levelId in body.LevelIds)
+        {
+            GameLevel? level = dataContext.Database.GetLevelById(levelId);
+            if (level == null) continue;
+
+            dataContext.Database.SetPlaylistLevelIndex(playlist, level, index);
+            index++;
+        }
```
jvyden, line 136: Any way to batch this? I could see this being pretty slow.

Refresh.GameServer/Endpoints/Game/Playlists/Lbp3PlaylistEndpoints.cs
```diff
@@ -0,0 +1,224 @@
...
+
+        (int skip, int count) = context.GetPageData();
+        IEnumerable<GamePlaylist> playlists = dataContext.Database.GetPlaylistsByAuthor(user);
+
+        return new SerializedLbp3PlaylistList 
+        (
+            SerializedLbp3Playlist.FromOldList(playlists.Skip(skip).Take(count), dataContext),
+            playlists.Count(),
+            skip
+        );
```
jvyden, line 178: 1. Client-side pagination should be moved to be a database query
2. If I understand it correctly, calling `.Count()` is forcing another enumeration so the db query will be executed twice which is even worse. There should be a dedicated method to counting the amount of playlists by the author.

Refresh.GameServer/Endpoints/Game/Playlists/Lbp3PlaylistEndpoints.cs
```diff
@@ -0,0 +1,224 @@
...
+
+        (int skip, int count) = context.GetPageData();
+        IEnumerable<GamePlaylist> playlists = dataContext.Database.GetPlaylistsFavouritedByUser(user);
+
+        return new SerializedLbp3FavouritePlaylistList
+        (
+            SerializedLbp3Playlist.FromOldList(playlists.Skip(skip).Take(count), dataContext),
+            playlists.Count(),
+            skip
+        );
```
jvyden, line 198: Ditto, db queries need to be rethought

Refresh.GameServer/Types/GameLocation.cs
```diff
@@ -22,4 +22,15 @@ public GameLocation(int x, int y)
...
     [XmlElement("x")] public int Y { get; set; }
+
+    public static GameLocation RandomLocation() 
+    {
+        Random random = new();
+        
+        return new GameLocation
+        (
+            random.Next(65535), 
+            random.Next(65535)
```
jvyden, line 33: Oh, if the max is known to be 65535 its probably worth moving GameLocation to use `ushort` instead. Unrelated though.

Also, you can use `Random.Shared` instead of making a new instance of `Random` every time

Refresh.GameServer/Types/Relations/FavouritePlaylistRelation.cs
```diff
@@ -0,0 +1,12 @@
...
+using Refresh.GameServer.Types.Playlists;
+using Refresh.GameServer.Types.UserData;
+
+namespace Refresh.GameServer.Types.Relations;
+#nullable disable
+
+public partial class FavouritePlaylistRelation : IRealmObject
+{
+    public GamePlaylist Playlist { get; set; }
+    public GameUser User { get; set; }
```
jvyden, line 11: It's generally good practice to store a timestamp of when these were added.

Refresh.GameServer/Database/GameDatabaseContext.Playlists.cs
```diff
@@ -101,56 +170,165 @@ public void AddLevelToPlaylist(GameLevel level, GamePlaylist parent)
...
         // TODO: with postgres this can be IQueryable
         => this.SubPlaylistRelations.Where(p => p.SubPlaylist == playlist).AsEnumerable()
+            .Reverse()
             .Select(r => this.GamePlaylists.First(p => p.PlaylistId == r.Playlist.PlaylistId))
             .Where(p => !p.IsRoot);
 
     public IEnumerable<GamePlaylist> GetPlaylistsByAuthorContainingPlaylist(GameUser user, GamePlaylist playlist)
         // TODO: with postgres this can be IQueryable
         => this.SubPlaylistRelations.Where(p => p.SubPlaylist == playlist).AsEnumerable()
+            .Reverse()
```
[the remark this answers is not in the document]
Toastbrot236: SubPlaylistRelations dont have a date, and reversing their order has them show up by recency first (this was also tested), an alternative would be ordering the playlists after the select

Refresh.GameServer/Database/GameDatabaseContext.Playlists.cs
```diff
@@ -101,56 +170,165 @@ public void AddLevelToPlaylist(GameLevel level, GamePlaylist parent)
...
         // TODO: When we have postgres, remove the `AsEnumerable` call for performance. 
         => this.LevelPlaylistRelations.Where(p => p.Level == level).AsEnumerable()
+            .Reverse()
             .Select(r => this.GamePlaylists.First(p => p.PlaylistId == r.Playlist.PlaylistId));
+
+    public IEnumerable<GamePlaylist> GetNewestPlaylists()
+        // TODO: When we have postgres, remove the `AsEnumerable` call for performance. 
+        => this.GamePlaylists.AsEnumerable()
+            .Where(p => !p.IsRoot)
+            .OrderByDescending(p => p.CreationDate);
```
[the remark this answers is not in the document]
Toastbrot236: i was told to put that comment everywhere i used AsEnumerable

Refresh.GameServer/Types/GameLocation.cs
Toastbrot236: that might actually make a lot of things easier, ill try it

Refresh.GameServer/Types/Relations/FavouritePlaylistRelation.cs
Toastbrot236: then i think it might be also good to add those timestamps to similar classes like FavouriteLevelRelation aswell?

Refresh.GameServer/Endpoints/Game/Playlists/Lbp3PlaylistEndpoints.cs
Toastbrot236: what do you mean with the first one?

Refresh.GameServer/Endpoints/Game/Playlists/Lbp3PlaylistEndpoints.cs
Toastbrot236: Would it be faster to have a database method for getting multiple levels by a list of ids? Ive seen some websites suggest using a combination of Where and Contains for similar things.

Refresh.GameServer/Database/GameDatabaseContext.Playlists.cs
```diff
@@ -1,16 +1,34 @@
...
 
 namespace Refresh.GameServer.Database;
 
 public partial class GameDatabaseContext // Playlists
 {
-    public GamePlaylist CreatePlaylist(GameUser user, SerializedPlaylist createInfo, bool rootPlaylist)
+    /// <summary>
+    /// Default icon used by playlists created in LBP3, through ApiV3 or similar
+    /// </summary>
+    private readonly string defaultPlaylistIcon = "g18451"; // LBP1 star sticker
```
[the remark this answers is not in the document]
Toastbrot236: vs code suggestions

Refresh.GameServer/Types/GameLocation.cs
Toastbrot236: Does Realm support ushort fields though?

Refresh.GameServer/Database/GameDatabaseContext.Playlists.cs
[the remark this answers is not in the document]
Toastbrot236: Now that I think about it more, ordering the playlists after the Select method is not very practical. My idea was that for example for levels/playlists in a playlist, they would be ordered by most recently added one first. I think just adding a timestamp to every relation should do it.

Refresh.GameServer/Endpoints/Game/Playlists/Lbp3PlaylistEndpoints.cs
jvyden: Yeah, I think that could be good.

Refresh.GameServer/Endpoints/Game/Playlists/Lbp3PlaylistEndpoints.cs
jvyden: Actually, I just realized. You should never return IEnumerable from a database function, try to return DatabaseList instead, which handles pagination and counts for you.

Refresh.GameServer/Types/GameLocation.cs
jvyden: Technically yes, technically no. It will compile, but internally in Realm it will be represented by a long.

Just ignore that fact for now. It will actually be ushort when postgres drops.

Refresh.GameServer/Endpoints/Game/Playlists/Lbp3PlaylistEndpoints.cs
Toastbrot236: Alright, makes sense

Refresh.GameServer/Types/GameLocation.cs
Toastbrot236: Alright, good to know

Refresh.GameServer/Endpoints/Game/Playlists/Lbp3PlaylistEndpoints.cs
Toastbrot236: I ended up doing a different method which doesn't use GameLevels at all and might generally be more efficient. I first wanted to quickly describe what it does here, but that was too much text so just look up UpdatePlaylistLevelOrder in GameDatabaseContext.Playlists

Refresh.GameServer/Types/GameLocation.cs
Toastbrot236: Now I've tried changing LocationX and LocationY of GamePlaylist, GameLevel and GameUser to ushort and when trying to compile, dotnet throws errors like: 
` error RLM024: GameLevel.LocationX is of type 'ushort' which is not yet supported. If that is supposed to be a model class, make sure it implements IRealmObject/IEmbeddedObject/IAsymmetricObject. ` 
(this is my first time trying code blocks in Github, wouldn't be surprised if it fails)

jvyden: Hmm... I see, just keep it as-is then.

Refresh.GameServer/Endpoints/Game/Playlists/Lbp3PlaylistEndpoints.cs
jvyden: I don't mind long comments; infact I prefer them if you think they're necessary. Will take a look though

Refresh.GameServer/Types/GameLocation.cs
```diff
@@ -5,12 +5,18 @@ namespace Refresh.GameServer.Types;
 [XmlType("location")]
 public class GameLocation
 {
-    public static GameLocation Zero => new()
+    public static GameLocation ZeroLocation => new()
     {
         X = 0,
         Y = 0,
     };
```
jvyden, line 12: Not a fan of this rename. BCL has Vector3.Zero for example, should keep for consistency

Refresh.GameServer/Types/GameLocation.cs
```diff
@@ -5,12 +5,18 @@ namespace Refresh.GameServer.Types;
...
 
+    public static GameLocation RandomLocation => new()
```
jvyden, line 14: Same here, just Random is verbose enough and consistent with BCL

jvyden (changes requested):
some additional things

Refresh.GameServer/Types/GameLocation.cs
jvyden: The Base Class Library, e.g. the base APIs that you use when you use .NET.

jvyden (approved):
Seems to work pretty great ingame! Just bugs with LBP3, really.

jvyden:
Actually wait I was supposed to test LBP1 as well. Hold on.

jvyden:
LBP1 seems to work too, merging

Issue #409, opened 2024-04-18
Ability to re-attribute levels to other users


jvyden:
Sometimes, users re-upload content from other instances, and then the original author comes onto our instance requesting that level be taken down.

It would be good to be able to swap a level's publisher so the level doesn't have to be deleted and all stats lost.

Would have been useful here: https://discord.com/channels/1049223665243389953/1049223665998372947/1230339741447356456

acidiclight:
I'd like to work on this, though I do need some help familiarizing myself with the codebase since I've literally never touched it.

jvyden:
Cool beans, I've assigned you then. o7

PR #722, opened 2025-03-07
Add API support for changing a level's author as an admin

acidiclight:
This merge request adds an admin endpoint for changing a level's author. You should be able to do this:

```
POST admin/levels/id/12345/setAuthor

{
    "authorName": "ritchie"
}
```

This would change the level with ID `12345`'s author to the user named `ritchie`. Both the level and user must actually exist in the database, otherwise you'll get a 404.

If you try to change the author of a level that has an `OriginalAuthor` value set on it, this value gets unset. This effectively addresses issue #409 .

I haven't actually tested this, I don't have a game patched and adding UI for this in refresh-web isn't my skillset. But it compiles. What could go wrong?

### DIFF
--- a/Refresh.GameServer/Database/GameDatabaseContext.Levels.cs
+++ b/Refresh.GameServer/Database/GameDatabaseContext.Levels.cs
@@ -65,6 +65,24 @@ public partial class GameDatabaseContext // Levels
         return level;
     }
 
+    public GameLevel UpdateLevelPublisher(GameLevel level, GameUser newAuthor)
+    {
+        // No need to change the level author if we haven't actually changed the author.
+        if (level.Publisher?.UserId == newAuthor.UserId)
+            return level;
+        
+        // this comment is not outdated.
+        this.Write(() =>
+        {
+            // Change the level's publisher, making sure we also unset OriginalPublisher
+            // if this level wasn't uploaded by an actual user originally.
+            level.Publisher = newAuthor;
+            level.OriginalPublisher = null;
+        });
+
+        return level;
+    }
+    
     public GameLevel? UpdateLevel(GameLevel newLevel, GameUser author)
     {
         if (newLevel.Title is { Length: > UgcLimits.TitleLimit })
--- a/Refresh.GameServer/Database/GameDatabaseContext.Levels.cs
+++ b/Refresh.GameServer/Database/GameDatabaseContext.Levels.cs
@@ -67,11 +67,9 @@ public partial class GameDatabaseContext // Levels
 
     public GameLevel UpdateLevelPublisher(GameLevel level, GameUser newAuthor)
     {
-        // No need to change the level author if we haven't actually changed the author.
         if (level.Publisher?.UserId == newAuthor.UserId)
             return level;
         
-        // this comment is not outdated.
         this.Write(() =>
         {
             // Change the level's publisher, making sure we also unset OriginalPublisher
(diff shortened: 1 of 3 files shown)


Refresh.GameServer/Database/GameDatabaseContext.Levels.cs
```diff
@@ -65,6 +65,24 @@ public GameLevel GetStoryLevelById(int id)
         return level;
     }
 
+    public GameLevel UpdateLevelPublisher(GameLevel level, GameUser newAuthor)
+    {
+        // No need to change the level author if we haven't actually changed the author.
+        if (level.Publisher?.UserId == newAuthor.UserId)
+            return level;
+        
+        // this comment is not outdated.
```
jvyden, line 74: What does this comment mean?

Refresh.GameServer/Endpoints/ApiV3/Admin/AdminLevelApiEndpoints.cs
```diff
@@ -72,4 +73,22 @@ public ApiOkResponse DeleteLevel(RequestContext context, GameDatabaseContext dat
...
         return new ApiOkResponse();
     }
+    
+    [ApiV3Endpoint("admin/levels/id/{id}/setAuthor", HttpMethods.Post), MinimumRole(GameUserRole.Curator)]
+    [DocSummary("Changes the author of a level. The new author must be an existing user on the server..")]
+    [DocError(typeof(ApiNotFoundError), ApiNotFoundError.LevelMissingErrorWhen)]
+    [DocError(typeof(ApiNotFoundError), ApiNotFoundError.UserMissingErrorWhen)]
+    public ApiOkResponse SetLevelAuthor(RequestContext context, GameDatabaseContext database, GameUser user, int id, ApiSetLevelAuthorRequest body)
+    {
+        var level = database.GetLevelById(id);
```
jvyden, line 83: We generally avoid `var` in Refresh.

Refresh.GameServer/Endpoints/ApiV3/DataTypes/Request/ApiSetLevelAuthorRequest.cs
```diff
@@ -0,0 +1,7 @@
+namespace Refresh.GameServer.Endpoints.ApiV3.DataTypes.Request;
+
+[JsonObject(NamingStrategyType = typeof(CamelCaseNamingStrategy))]
+public sealed class ApiSetLevelAuthorRequest
+{
+    public string AuthorName { get; set; } = string.Empty;
```
jvyden, line 6: For requests like this we typically use user IDs.

jvyden (changes requested):

Refresh.GameServer/Database/GameDatabaseContext.Levels.cs
acidiclight: Tired ritchie way of saying there's no reason to actually write to the database if the user ID isn't actually going to change.

Refresh.GameServer/Endpoints/ApiV3/Admin/AdminLevelApiEndpoints.cs
acidiclight: Sorry about that; the Rider project config isn't set to warn about that.

Refresh.GameServer/Endpoints/ApiV3/DataTypes/Request/ApiSetLevelAuthorRequest.cs
acidiclight: Couldn't find a way to get a user by ID from the database. There was a `GetUserByObjectId` but it takes an `ObjectId` rather than `int`. It wasn't clear to me how I'd go about accepting that in the request body, so I went with username just to be safe.

jvyden (approved):

PR #732, opened 2025-03-18
Include level type in GameMinimalLevelResponse

### LINKED ISSUE #397: The special badges for different gamemodes don't show up on the earth
If a level is set as a VS or cutscene level, the only way of finding this out right now is either by pressing Square to find out more or playing the level in question

this issue partially affects the level browser aswell - the gamemode overlay will show until the level creator's earth is loaded

Toastbrot236:
Fixes that bug where versus/cutscene levels would only show their badges in-game in certain cases

Closes #397


Toastbrot236:
fixes issue #397

jvyden (approved):

PR #738, opened 2025-03-19
Add solution configuration for Postgres implementation development

jvyden:
This introduces a new `DebugPostgres` configuration in the solution to facilitate working on Postgres on the main development branch. The idea is that this can be worked on iteratively, while not having the branch rot like the previous two efforts of working on Postgres.

I also did some work as an example on how Realm and Postgres should co-exist, leveraging ifdefs and `global using`s to prevent too much ifdef spam at the top of files. Unfortunately I do have to duplicate the list of `DbSet`s in `GameDatabaseContext`. I tried defining `DbSet` as `RealmDbSet<T>` to mitigate this but C#'s preprocessor can't handle generics.

## The key component here is that changes on this branch are not required to compile.
That sounds weird I know, but this PR is certainly **not** a full implementation, merely a couple steps in that direction. This is so relatively small PRs can be made. Obviously, changes should make sense but we're not going to have this compiling in one commit.

Most of the current database code seems to compile, with the exclusion being calls to `AddSequentialObject`, and objects that inherit `IRealmObject` (that interface doesn't exist)

Aside from those concerns, this should hopefully be the last attempt to get Postgres kicked off.

### DIFF
--- a/Refresh.GameServer/Database/GameDatabaseContext.Relations.cs
+++ b/Refresh.GameServer/Database/GameDatabaseContext.Relations.cs
@@ -493,7 +493,13 @@ public partial class GameDatabaseContext // Relations
     public int GetTotalRatingsForLevelComment(GameLevelComment comment, RatingType type) =>
         this.LevelCommentRelations.Count(r => r.Comment == comment && r._RatingType == (int)type);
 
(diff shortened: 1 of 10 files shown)


jvyden:
Nobody seems to be interested in reviewing this so I'm going to reclaim my merge-without-review rights.

PR #751, opened 2025-04-17
Pin refactor

Toastbrot236:
This PR reworks pins to be stored using seperate relation classes (`PinProgressRelation` and `ProfilePinRelation`) instead of an embedded class in `GameUser`. This makes pins hopefully generally easier to handle and prepares for future feature implementations, like awarding pins manually through the server and ApiV3 support for pins. Profile pins can be set per game, and pin progress is cross-game, only split between Beta build and Not beta build.


Refresh.GameServer/Database/GameDatabaseContext.Pins.cs
```diff
@@ -0,0 +1,115 @@
...
+                }
+            }
+        });
+    }
+
+    public void UpdateUserProfilePins(List<long> profilePins, GameUser user, TokenGame game)
+    {
+        DateTimeOffset now = this._time.Now;
+        IEnumerable<PinProgressRelation> exProgresses = this.GetPinProgressesByUser(user, game == TokenGame.BetaBuild);
+        IEnumerable<ProfilePinRelation> exProfilePins = this.GetProfilePinsByUser(user, game);
```
jvyden, line 48: `ex`?

Refresh.GameServer/Endpoints/Game/UserEndpoints.cs
```diff
@@ -146,38 +147,51 @@ public SerializedFriendsList GetFriends(RequestContext context, GameDatabaseCont
...
-        //Dont serialize profile pins, the packet capture doesnt have them in the return
-        updateUserPins.ProfilePins?.Clear();
+        if (pinProgresses.Count > 0)
+        {
+            dataContext.Database.UpdateUserPinProgress(pinProgresses, user, dataContext.Game);
+        }
 
-        //Just return the same pins back to the client
-        return JsonConvert.SerializeObject(updateUserPins, new JsonSerializerSettings
+        if (body.ProfilePins.Count > 0)
```
jvyden, line 179: Should add validation for max of 3 pins

jvyden (changes requested):

Refresh.GameServer/Database/GameDatabaseContext.Pins.cs
Toastbrot236: It was meant to be an acronym for "existing", but now i see it might be confusing, since its explained nowhere. Maybe "old" might be a better prefix?

Refresh.GameServer/Endpoints/Game/UserEndpoints.cs
Toastbrot236: Actual validation where lists with more than 3 pins get rejected, or just taking the first 3 pins from the list?

Refresh.GameServer/Database/GameDatabaseContext.Pins.cs
Toastbrot236: I ended up using "existing" since that is more accurate I think

Refresh.GameServer/Endpoints/Game/UserEndpoints.cs
Toastbrot236: I did the latter for now

jvyden (approved):

PR #776, opened 2025-05-22
Add validation for patchwork useragents

jvyden:
Opening as draft as versioning scheme isn't finalized yet.

### DIFF
--- a/Refresh.Common/Extensions/RequestContextExtensions.cs
+++ b/Refresh.Common/Extensions/RequestContextExtensions.cs
@@ -27,6 +27,43 @@ public static class RequestContextExtensions
 
     [Pure]
     public static bool IsApi(this RequestContext context) => context.Url.AbsolutePath.StartsWith("/api/");
+
+    [Pure]
+    public static bool IsPatchworkVersionValid(this RequestContext context, int requiredMajor, int requiredMinor)
+    {
+        ReadOnlySpan<char> userAgent = context.RequestHeaders.Get("User-Agent").AsSpan();
+        if (userAgent.IsEmpty)
+            return false;
+
+        return IsPatchworkVersionValid(userAgent, requiredMajor, requiredMinor);
+    }
+
+    [Pure]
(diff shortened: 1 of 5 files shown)


Refresh.GameServer/Configuration/GameServerConfig.cs
```diff
@@ -64,6 +64,21 @@ protected override void Migrate(int oldVer, dynamic oldConfig)
...
+
+    /// <summary>
+    /// The minimum required major version of Patchwork on the client to be able to connect.
+    /// </summary>
+    public int RequiredPatchworkMajorVersion { get; set; } = 1;
+    
+    /// <summary>
+    /// The minimum required minor version of Patchwork on the client to be able to connect.
+    /// </summary>
+    public int RequiredPatchworkMinorVersion { get; set; } = 0;
```
jvyden, line 81: We would ideally add migrations to automatically increase this when Patchwork updates and increases security.

To avoid enforcing everyone to repatch all the time we should avoid increasing this whenever possible.

Refresh.Common/Extensions/RequestContextExtensions.cs
```diff
@@ -27,6 +27,43 @@ public static (int, int) GetPageData(this RequestContext context, int maxCount =
...
+        if (!userAgent.StartsWith(namePrefix))
+            return false;
+        
+        // is the game version valid?
+        char gameVersion = userAgent[namePrefix.Length];
+        if (gameVersion is not '1' and not '2' and not '3' and not 'V')
+            return false;
+
+        // does the version string parse out?
+        if (!Version.TryParse(userAgent[(namePrefix.Length + 2)..], out Version? version))
```
jvyden, line 57: This parsing is flexible and should hold up if we want to add patch versions (e.g. 1.0.x)

Zaprit (approved):

jvyden:
Merging as versioning has been hashed out internally.

PR #784, opened 2025-06-01
Fix some warnings

Toastbrot236:
This fixes some warnings which VS Code has now started to treat as errors.


jvyden (approved):
Interesting that this doesn't show up in CI or Rider

PR #798, opened 2025-06-11
Get Postgres compiling!

jvyden:
I wrote a bunch of polyfills for Realm attributes, and vice-versa for Postgres and Postgres is now compiling!

This requires a hacky workaround that requires us to specify two attributes every time we require one, but this is such a big step in the right direction.

It runs, but doesn't get very far since we're missing actual initialization code:
image


jvyden:
Nothing seems to be broken on Realm so I'm gonna go ahead and merge this!

PR #843, opened 2025-06-29
Truncate level publisher names to fix game not showing some fake publishers

### LINKED ISSUE #698: Fake User's username doesn't show up in-game if it's 16 characters long
in other words, when adding the ! to the username, it makes it 17 characters long instead of 16, which the game outright refuses to display (as pictured below):
image
a simple fix would be to cut off the last letter of the username, (and maybe 3 more to add an elipsis at the end of said username)
https://lbp.littlebigrefresh.com/level/6175 here's the level on the website for comparison

Nya0:
Fixes #698

### DIFF
--- a/Refresh.Core/Refresh.Core.csproj
+++ b/Refresh.Core/Refresh.Core.csproj
@@ -22,7 +22,7 @@
         <PackageReference Include="CommandLineParser" Version="2.9.1" />
         <PackageReference Include="Discord.Net.Webhook" Version="3.17.4" />
         <PackageReference Include="DnsClient" Version="1.8.0" />
-        <PackageReference Include="MailKit" Version="4.12.1" />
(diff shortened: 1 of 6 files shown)


jvyden (approved):
Looks okay, though I'd like to test once I'm home to make sure.

What games have you tested against for reference? In particular I could see PSP not being able to render that character but that's just a hunch.

Nya0:
Ok it doesnt seem to work becauase while ellipis U+2026 `…` is 1 char. it counts as 3 bytes so basically it wouldnt work and we are better off using ...

Beyley:
> I could see PSP not being able to render that character but that's just a hunch.

From what I remember, PSP will have issues with special Unicode characters like these. JP text and some non-ASCII European languages are fine, but others (like special punctuation or most all "weird" Unicode characters) are either missing chars in the fonts or won't parse at all.

Issue #654, opened 2024-09-05
Recent Activity Events for commenting on a level/profile


Arcadius2006:
basically #648 but i used the correct terminology this time so i don't accidentally suggest an admittedly pretty good feature that i wasn't even thinking of at the time

this suggestion is also pretty similar to #476, but i don't know if these work similarly internally or not, so idk

Nya0:
is this not completed by #653?

Nya0:
shouldn't both be the same for the sake of consistency?

game events or website notifications both should be in recent activity

Arcadius2006:
yknow ur right you should try implementing that

Issue #476, opened 2024-05-10
Implement Recent Activity notification for leaving a review on someone's level


Arcadius2006:
LBP2 and LBP3 notified you if someone has reviewed one of your levels in recent activity

this functionality doesn't seem to be present in refresh, however

PR #850, opened 2025-07-03
web notifications and game events for comments/reviews

Nya0:
#654 #476 #648

### DIFF
--- a/Refresh.Database/GameDatabaseContext.Activity.cs
+++ b/Refresh.Database/GameDatabaseContext.Activity.cs
@@ -109,6 +109,7 @@ public partial class GameDatabaseContext // Activity
         {
             List<ObjectId?> favouriteUsers = this.GetUsersFavouritedByUser(parameters.User, 1000, 0).Select(f => (ObjectId?)f.UserId).ToList();
             List<ObjectId?> userFriends = this.GetUsersMutuals(parameters.User).Select(u => (ObjectId?)u.UserId).ToList();
+            List<int> userLevels = this.GetLevelsByUser(parameters.User).Select(l => l.LevelId).ToList();
             
             query = query.Where(e =>
                 e.User?.UserId == parameters.User.UserId ||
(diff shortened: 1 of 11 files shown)


Nya0:
oh wait these are not web notifications im so smart

Refresh.Database/GameDatabaseContext.Activity.cs
```diff
@@ -52,7 +52,7 @@ private IEnumerable<Event> GetEvents(ActivityQueryParameters parameters)
 
         if (parameters is { ExcludeMyself: true, User: not null })
         {
-            query = query.Where(e => e.User.UserId != parameters.User.UserId && (e.StoredDataType != EventDataType.User || e.StoredObjectId != parameters.User.UserId));  
+            query = query.Where(e => e.User != parameters.User);
```
jvyden, line 55: Is this an intentional behavior change?

Refresh.Database/GameDatabaseContext.Levels.cs
```diff
@@ -295,6 +295,7 @@ private IQueryable<GameLevel> GetLevelsByGameVersion(TokenGame gameVersion)
             .FilterByGameVersion(gameVersion);
 
     [Pure]
+    public IQueryable<GameLevel> GetLevelsByUser(GameUser user) => this.GameLevels.Where(l => l.Publisher == user);
```
jvyden, line 298: I think this should be named GetAllLevelsByUser to avoid overloading the paginated query that has different behavior

jvyden (commented):
I understand this is a draft so these are just a couple notes.

I would consider game-specific notifications to be out of scope for this PR. I've already mentally drafted how those will work so please wait for that infrastructure to exist for now.

Refresh.Database/GameDatabaseContext.Levels.cs
Nya0: good call although when i was trying to use it it never occurred to me it will require anything more than a user instance

so i think its more natural for the complex option that includes filtering to have a more fit name like `GetLevelsByUserFiltered`

Nya0:
do u think u will be free able to talk about it over discord? much appreciated

jvyden:
Regarding what?

Nya0:
is there anything else like i should know like any other changes u already have planned

Nya0:
@jvyden any notes? appreciated

Nya0:
adding #582 to pool for this pull request

jvyden:
Please don't. Keep PRs focused to one thing at a time.

jvyden:
I don't really know what's happening with this PR, so I'm gonna close for now.

You can use `git cherry-pick` to extract changes out from this branch if you wish, but as-is I don't accept PRs that change multiple random things at once.

PR #856, opened 2025-07-07
Playlist fixes post-database migration

Toastbrot236:
This fixes everything playlist-related that broke after the Postgres migration, while also cleaning up `GameDatabaseContext.Playlist` in order to achieve that. Also, `GameUser.RootPlaylist` had to be removed, since both it and `GamePlaylist.Publisher` led to EF having a one-to-one relationship between users and playlists, preventing users from creating more than one playlist, and on top of that it was redundant information.

### DIFF
--- a/Refresh.Database/GameDatabaseContext.Playlists.cs
+++ b/Refresh.Database/GameDatabaseContext.Playlists.cs
@@ -159,60 +159,50 @@ public partial class GameDatabaseContext // Playlists
         if (relation == null)
             return;
 
-        // decrease index of every playlist level after this one by 1
-        this.DecreasePlaylistLevelIndicesAfterIndex(parent, relation.Index);
-
         this.Write(() =>
         {
             this.LevelPlaylistRelations.Remove(relation);
         });
     }
 
-    private void DecreasePlaylistLevelIndicesAfterIndex(GamePlaylist playlist, int startIndex)
+    public void ReorderLevelsInPlaylist(IEnumerable<int> levelIds, GamePlaylist parent)
     {
-        IEnumerable<LevelPlaylistRelation> relations = this.LevelPlaylistRelations
-            .Where(r => r.Playlist == playlist && r.Index >= startIndex)
-            .AsEnumerable();
+        IEnumerable<LevelPlaylistRelation> relations = this.GetLevelRelationsForPlaylist(parent).ToArray();
+        IEnumerable<LevelPlaylistRelation> includedRelations = relations.Where(r => levelIds.Contains(r.LevelId));
+        IEnumerable<LevelPlaylistRelation> excludedRelations = relations.Where(r => !levelIds.Contains(r.LevelId));
 
-        this.Write(() => {
-            foreach(LevelPlaylistRelation relation in relations)
-            {
-                relation.Index--;
-            }
-        });
-    }
-
-    public void UpdatePlaylistLevelOrder(GamePlaylist playlist, List<int> levelIds)
-    {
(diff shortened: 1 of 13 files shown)


jvyden:
Tests failing

Toastbrot236:
After many hours, I've finally got local tests working. Now I'll just have to find out why these tests are failing. They all seem to be about the `GetFavouriteUsers` endpoint, and I still have to find out what that has to do with playlists.

Refresh.Database/GameDatabaseContext.Playlists.cs
```diff
@@ -154,179 +165,128 @@ public void AddLevelToPlaylist(GameLevel level, GamePlaylist parent)
...
 
-    public int GetTotalLevelsInPlaylistCount(GamePlaylist playlist) => 
-        this.LevelPlaylistRelations.Count(l => l.Playlist == playlist);
+    public int GetTotalLevelsInPlaylistCount(GamePlaylist playlist) 
+        => this.LevelPlaylistRelations.Count(l => l.Playlist == playlist);
 
-    public IEnumerable<GamePlaylist> GetPlaylistsInPlaylist(GamePlaylist playlist)
-        // TODO: When we have postgres, remove the `AsEnumerable` call for performance. 
+    [Obsolete("Only to be used by GameDatabaseContext and GamePlaylistExtensions.")]
+    public IEnumerable<GamePlaylist> GetPlaylistsContainingPlaylist(GamePlaylist playlist)
```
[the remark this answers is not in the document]
jvyden: Hmm... I don't see why you couldn't just move the extension class into `Refresh.Database`, where it can access the `internal` methods but remain public to consumers.

I could also see the method being moved into `GamePlaylist` directly but I'd prefer to keep code out of the models.

Refresh.Interfaces.Game/Endpoints/DataTypes/Response/GameUserResponse.cs
```diff
@@ -107,7 +105,10 @@ public class GameUserResponse : IDataConvertableFrom<GameUserResponse, GameUser>
             
             return response;
         }
-        
+
+        response.RootPlaylistId = dataContext.Database.GetUserRootPlaylist(old)?.PlaylistId.ToString();
```
jvyden, line 109: Could we check if the game is LBP1 before setting this?

Refresh.Interfaces.Game/Endpoints/DataTypes/Response/GameUserResponse.cs
```diff
@@ -107,7 +105,10 @@ public class GameUserResponse : IDataConvertableFrom<GameUserResponse, GameUser>
...
+        response.ProfilePins = dataContext.Database.GetProfilePinsByUser(old, dataContext.Game, 0, 3).Items.Select(p => p.PinId).ToList();
```
jvyden, line 110: Same here, check if the game *isn't* LBP1/PSP

jvyden (commented):
Minor things. Sorry for the late review, I've been helping test Resonite prereleases.

Refresh.Interfaces.Game/Endpoints/DataTypes/Response/GameUserResponse.cs
Toastbrot236: If we already do game-checking for these two attributes, why not do it for all other attributes in GameUserResponse aswell? Things like ReviewCount, YayIconHash, QueuedLevelCount etc

jvyden: 100%, we should. but just to keep things concise let's do that in a follow-up PR.

jvyden (approved):

PR #858, opened 2025-07-09
Add helper method for extracting attributes from level descriptions

### LINKED ISSUE #791: Ability to alter levels' original upload dates
Not sure if this is a server or refresh web issue (probably both?), apologies if this is misplaced

Purpose would be three-fold:

1. from an archival standpoint it'd be cool to be able to see levels original publish date without use of an external source

2. would be a good way to help denote levels that are 'remixes' but not quite exact reuploads (ie restorative work/bug fixes done by a reuploader)

and most importantly

3. send reuploaded levels to their proper upload date on the in-game/website 'newest levels' timeline (which, for most reuploaded levels would send them to the furthest back, a time before this server was even up).
This would be a great but fair way to deter reuploads and fix the newest category so that it highlights actual new levels

Nya0:
~Closes #791~ ~(although we should maybe only close it after website implementationPS: #859)~

ps: small improvements to attribute system

### DIFF
--- a/RefreshTests.GameServer/Tests/Levels/ReuploadDetectionTests.cs
+++ b/RefreshTests.GameServer/Tests/Levels/ReuploadDetectionTests.cs
@@ -1,4 +1,5 @@
-﻿using Refresh.Database.Models.Levels;
+﻿using Refresh.Common.Constants;
+using Refresh.Database.Models.Levels;
 using Refresh.Database.Models.Users;
 
 namespace RefreshTests.GameServer.Tests.Levels;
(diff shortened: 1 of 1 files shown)


jvyden (changes requested):
I'm not sure if we should expose this to users, or if we even want this feature at all... I kind of want to wait to just do the big import before adding things like this.

If we do end up doing this, we should at least make this a new field, but right now I'm kind of leaning towards 'no'.

@CrankyLandlord Have the new cool levels changes and the ability to hide reuploads nullified the point of your original issue?

jvyden (changes requested):
Tests failing

Nya0:
forgot to remove tests i added

PR #864, opened 2025-07-13
Add migrations to bump level ids by 200,000,000

jvyden:
Some preliminary work for the big archive import. This will allow us to cleanly import everything while keeping all the old level ids from the original servers.

The number 200,000,000 was chosen because it's about double the amount of the highest ID recorded in the dry database, that number being 100,515,565.

The sooner we get this out, the lesser the impact this will have as people continue to upload levels.

Marking as draft because there are a few considerations:
- How do we handle website links? Every level link ever sent will break with this update.
   - We could introduce a new level route for the website, and redirect the old one, for example `https://lbp.lbpbonsai.com/l/1234`. There's also the old LBP.me links, like `https://lbp.me/v/<some encoded int>`.
   - We could just redirect and add 200,000,000 for now to keep it simple.
- Are the new foreign key definitions correct? Will it cause unexpected behavior?
- There may be some cases where the Level ID isn't referenced by a foreign key. An example of this is Events, but there could be more of these that I'm unaware of.

### DIFF
--- a/Refresh.Database/Migrations/20250713213830_BumpLevelIds.cs
+++ b/Refresh.Database/Migrations/20250713213830_BumpLevelIds.cs
@@ -1,0 +1,64 @@
+﻿using Microsoft.EntityFrameworkCore.Infrastructure;
+using Microsoft.EntityFrameworkCore.Migrations;
+
+#nullable disable
+
+namespace Refresh.Database.Migrations
+{
+    /// <inheritdoc />
+    [Migration("20250713213830_BumpLevelIds")]
+    [DbContext(typeof(GameDatabaseContext))]
+    public partial class BumpLevelIds : Migration
+    {
+        private void AddForeignKey(MigrationBuilder migration, string name, string table, string column = "LevelId")
+        {
+            migration.Sql($"UPDATE \"{table}\" SET \"{column}\" = \"{column}\" + 200000000");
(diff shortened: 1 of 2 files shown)


jvyden:
True, but it also goes the other way... if we don't do this level keys & links will be broken in reuploaded levels.

I don't think there's much we can do about that, and I think most instances of keys are spam anyways. Level links are a bad thing to break though.

Nya0:
It would not be a bad idea to get some simple infrastructure for that. might also be useful in the future. it seems to be the best option cuz whats proposed right now just seems like duct tape that we might regret in the future

Nya0:
actually i dont know

Toastbrot236:
For the infrastructure, maybe we could look into [CwLibNet](https://github.com/ProjectCraftworld/CwLibNet/tree/main)? I don't know how feature complete it is exactly (aside from seemingly having code for level links/keys), maybe if that doesn't work we could probably use [CwLib](https://github.com/ennuo/toolkit/tree/cwlib) to write a small Java program to migrate the data store?

jvyden:
I won't go into detail here but I'd like to avoid that library since we (Bonsai) have had issues with the maintainers of that project.

I was actually thinking we could just consume toolkit in a Java server. We already plan to support moving workers outside of the main `GameServer`, so maybe that could be an addition to that - we could have the C# worker queue up levels to migrate, and have the Java side migrate them.

The downside to this is that it requires you to write Java.

PR #904, opened 2025-07-29
Implement Playlist Statistics

Toastbrot236:
Adds playlist statistics, which work similarly to level and user statistics. Also slightly optimizes some playlist database methods and groups a few of them together to hopefully make them easier to follow.


jvyden (changes requested):
With the statistics system, it's important to keep track of the versioning as it's how Refresh knows when to update objects. Without bumping them with schema changes they will be out of date and will never catch up unless marked dirty.

I think I implemented it so that Refresh serves old data while a worker updates everything in the background.

jvyden:
Very unexpected. Are they not being cascade deleted?

jvyden (approved):

Issue #700, opened 2024-12-07
LBP3 disconnects after a certain amount of time passes


Arcadius2006:
if you play LBP3 for long enough, the game will disconnect you from the server for no clear reason, making you have to connect again

jvyden:
The above is probably the case. As I understand it the game should be automatically grabbing a new token, though? Is this an error popping up in-game @Arcadius2006?

Nya0:
### LBP3 RPCS3
- Set `GameTokenExpirySeconds` to 1 minute for testing

The game successfully reauthenticates for the first 5 attempts:

![image](https://github.com/user-attachments/assets/23c88b0a-0f8a-4fb1-b12e-32096ebe1e5a)

After multiple successful reauthentications, the game stops attempting to reauthenticate for some reason:
- No error messages appear in-game
- Only loading indicators are shown

![image](https://github.com/user-attachments/assets/b81ae7c2-13cb-483d-a8e4-14a598f130e2)

But other times it just does it 10+ times before it stops working?

![image](https://github.com/user-attachments/assets/ba319b03-5609-4463-8929-f1ff4ccf9166)

Unsure what to conclude other than the game eventually just doesnt even try anymore. idk how the behavior is on PS3.

Nya0:
what i reached on my end did not look like a ratelimit from the server, it seems to be really random and more like a game bug 

this could be a separate issue i think

Toastbrot236:
The server rate limits the login requests for me because once the game starts spamming them, it does one per ~30 seconds. The server generally rate limits game login requests once a user exceeds 10 per 5 minutes. You're not hitting the rate limit because they're happening more rarely for you. Also for me the server never responds with 403, all requests which require authentication are successful.
So idk, but my guess is for you the game starts to send invalid tokens, leading to the server responding with 403 and the game trying to reauthenticate because of that (which LBP does for most requests if the server responds with 403 to them). For me the game for some reason tries to reauthenticate without receiving 403 prior. This is really weird.

Toastbrot236:
Maybe this could be related to Bosnai's responses not containing the `X-Original-Content-Length`, `Vary` and `Content-Encoding` headers, which `DeflateMiddleware` would normally add when non-asset game responses become too large? My instance seems to set them fine, but they're missing for Bonsai.

jvyden:
Could be a Cloudflare configuration error, or something happening during ingress.

I'll try and capture some traffic and see if the headers are there.

[...]

PR #908, opened 2025-07-30
Also check for gzip when considering request encodings

jvyden:
As discussed on Discord. https://discord.com/channels/1049223665243389953/1049225857350254632/1400079015754334259

Pretty much, we haven't been properly compressing some requests for a while, because some requests ask for `gzip` when they are really asking for `deflate`. I'm 99% sure LBP just sends bullshit for the `Accept-Encoding` header half the time.

Lets see how this goes. This may or may not require a revert so I'll try and do some testing first. I'm unsure if this would be LLE/HLE in RPCS3 so it would be good to test on both RPCS3 and PS3. PSP and Vita would also be good to test but I don't have access to those platforms.

Potential solution for #700.


jvyden:
I did some testing and this seems okay. YOLO

PR #912, opened 2025-07-31
Optimize some level responses

Toastbrot236:
This makes `GameLevelResponse.FromOld()` depend on `GameMinimalLevelResponse.FromOld()` instead of the other way, to avoid making database calls for information like the user's own review which will not be used for `GameMinimalLevelResponse` anyway, while having as little duplicate code in both methods as possible. Also a few database calls in these methods will only be done if the user is in a game which supports the attributes these calls are for (for example, the user's review won't be gotten if they are in LBP1 or PSP, and the level's skill rewards won't be gotten if the user is not in LBP Vita or a beta build). Also makes LBP3 playlist endpoints return `SerializedMinimalLevelList` instead of `SerializedLevelList`.


jvyden:
This seems to break LBP1.

PR #915, opened 2025-08-01
Fix GameLevelResponse icon setting

Toastbrot236:
Fixes a typo/regression I made where the `GameMinimalLevelResponse` and `GameLevelResponse` icon hashes would be left blank in some cases.


jvyden (approved):
Your honor, I call tee-hee

Issue #690, opened 2024-11-19
Implement missing pin awards


Arcadius2006:
some pin awards, like the one for getting at a certain scoreboard percentage or getting enough positive ratings on a level, are not handled by the game and instead expect to be awarded by the server instead. locate all the pins affected and try to implement them as closely as possible to their original functionality

Toastbrot236:
I have started to work on this recently alongside issues #674 and #395

Issue #395, opened 2024-04-07
Unlock LBP.me pins upon API sign-in


jvyden:
https://discord.com/channels/1049223665243389953/1049225857350254632/1226633636548317204

PR #918, opened 2025-08-03
Implement manual pin awarding and pin tests

Toastbrot236:
This PR:
- Adds database methods for awarding single `PinProgressRelation`s to users, aswell as `ManuallyAwardedPins`, which is an enum tracking the progressTypes of all pins (right now not all pins) which the game does not unlock itself and instead expects the server to award manually
- Implements awarding all 4 leaderboard-related pins aswell as 2 out of 3 website-related pins from `ManuallyAwardedPins` to users if they complete their objectives, as an example for the above
- Adds tests for the above
- Slightly adjusts pin progress syncing for 2 certain manually-awarded pins whose progress works differently from all other pins
- Adds a test for the above aswell

This would've closed both #395 and #690, however the pin awarded for hearting a user on the website is not implemented yet since there is no way to heart users using the API yet, and the rating pins are not implemented yet either because I didn't really want this PR to become too large.

Also my naming skills aren't very great.


Refresh.Database/GameDatabaseContext.Pins.cs
```diff
@@ -11,30 +13,43 @@ public void UpdateUserPinProgress(Dictionary<long, int> pinProgressUpdates, Game
         DateTimeOffset now = this._time.Now;
         bool isBeta = game == TokenGame.BetaBuild;
         IEnumerable<PinProgressRelation> existingProgresses = this.GetPinProgressesByUser(user, isBeta);
-       
+        FrozenSet<long> specialTreatmentPins =
+        [
+            (long)ManuallyAwardedPins.TopXOfAnyStoryLevelWithOver50Scores,
+            (long)ManuallyAwardedPins.TopXOfAnyCommunityLevelWithOver50Scores,
+        ];
```
jvyden, line 20: As far as I know it's a bit slow to freeze a list. This should be done statically at the class level.

For naming I think `progressDescPins`, or to be verbose, `pinsWithDescendingProgress`

Refresh.Interfaces.APIv3/Endpoints/AuthenticationApiEndpoints.cs
```diff
@@ -107,6 +108,10 @@ public ApiResponse<IApiAuthenticationResponse> Authenticate(RequestContext conte
         
         context.Logger.LogInfo(BunkumCategory.Authentication, $"{user} successfully logged in through the API");
 
+        // Update pin progress for signing into the API
+        database.IncrementUserPinProgress((long)ManuallyAwardedPins.SignIntoWebsite, 1, user, false);
+        database.IncrementUserPinProgress((long)ManuallyAwardedPins.SignIntoWebsite, 1, user, true);
```
jvyden, line 113: Bleh, this sucks. Maybe that last param should be nullable, where null means update for both beta, not-beta?

Refresh.Database/Models/Pins/ManuallyAwardedPins.cs
```diff
@@ -0,0 +1,18 @@
+namespace Refresh.Database.Models.Pins;
+
+/// <summary>
+/// The progress types of pins which have to be awarded manually by the server.
+/// </summary>
+public enum ManuallyAwardedPins : long
```
jvyden, line 6: `ServerPins` maybe?

Toastbrot236: I think that might make it sound like these are our own custom pins or something, but other than that I like that name.

Refresh.Database/GameDatabaseContext.Pins.cs
Toastbrot236: descendingProgressPins maybe? Is verbose but a little shorter.

jvyden (approved):

PR #925, opened 2025-08-09
Pin progress updating fixes and improvements

Toastbrot236:
This PR fixes manually awarded pins not having their progress properly incremented when updating them, extends the `AchieveTopFourthOfXLeaderboardsPin` test to test that case aswell, and does some minor improvements like replacing all `Write` calls in `GameDatabaseContext.Pins` with just `SaveChanges` calls, correcting the namespaces of pin tests and renaming the `ManuallyAwardedPins` file to match its class name.


Refresh.Database/GameDatabaseContext.Pins.cs
```diff
@@ -18,41 +18,39 @@ public void UpdateUserPinProgress(Dictionary<long, int> pinProgressUpdates, Game
...
+            // specialTreatmentPins, it's better the smaller it is.
+            if ((!isSpecialTreatmentPin && newProgress > existingProgress.Progress)
+                || (isSpecialTreatmentPin && newProgress < existingProgress.Progress))
+            {
+                existingProgress.Progress = newProgress;
+                existingProgress.LastUpdated = now;
             }
-        });
+        }
+        SaveChanges();
```
jvyden, line 53:
```suggestion
        this.SaveChanges();
```

and everywhere else it's used

RefreshTests.GameServer/TestContext.cs
```diff
@@ -147,7 +147,8 @@ public void FillLeaderboard(GameLevel level, int count, byte type)
     {
         for (byte i = 0; i < count; i++)
         {
-            GameUser scoreUser = this.CreateUser("score" + i);
+            string username = "score" + i;
+            GameUser scoreUser = Database.GetUserByUsername(username) ?? this.CreateUser(username);
```
jvyden, line 151:
```suggestion
            GameUser scoreUser = this.Database.GetUserByUsername(username) ?? this.CreateUser(username);
```

jvyden (changes requested):

jvyden (approved):

PR #926, opened 2025-08-12
Improve result filter initialization, add new filters

Toastbrot236:
This PR renames `LevelFilterSettings` to `ResultFilterSettings`, adds new filters to that class and implements most of them (those being lbp3-exclusive filters like adventure filtering and the result types to filter, which is why I renamed the class, aswell as filtering levels which require move controllers), and splits the constructor into static initialization classes, one for API requests and one for game requests. The new filters I added currently only work for game requests, not for API requests, to not make this PR too long.


Toastbrot236:
I'm doing the rebase right now

Toastbrot236:
Merging like this ended up being easier. Also, now that I've reverted the rename, I hope this PR is easier to follow now.

jvyden (approved):

PR #940, opened 2025-08-26
Don't try to block commands while uploading a level

### LINKED ISSUE #937: Commands don't work for certain people.
i was playing with Aether (PSN username `NEMO33`, discord handle `@atherlie_`), and i found it that he couldn't execute any commands entirely. here's a picture of this in action: 

Image

Toastbrot236:
This gets rid of the mechanism which uses `CommandService` to block commands while publishing a level.
 
We can't catch all cases of `/filter` requests which are not chat messages (e.g. LBP3 sends the level strings to filter before its `/startPublish` request, and LBP2 can also send level strings during create mode).
More importantly, if the game successfully does a `/startPublish` request, but its `/publish` request fails in any way or the user tries to update a level instead of publishing a new one, their commands would continue to silently get blocked, which has already caused confusion (see #937).

The latter issue could be fixed and worked around by making some changes to the `/publish` endpoint method and the `CommandService`, but since we also can't tell whether a `/filter` request does actually contain a chat message or something else half the time, making the mechanism not very effective to begin with, I think it makes more sense to just get rid of it entirely.

Closes #937.

### DIFF
--- a/Refresh.Core/Services/CommandService.cs
+++ b/Refresh.Core/Services/CommandService.cs
@@ -21,30 +21,6 @@ public class CommandService : EndpointService
         this._levelListService = levelListService;
     }
 
-    private readonly HashSet<ObjectId> _usersPublishing = [];
-
-    /// <summary>
-    /// Start tracking the user, eg. they started publishing
-    /// </summary>
-    /// <param name="id">The user ID</param>
-    public void StartPublishing(ObjectId id)
-    {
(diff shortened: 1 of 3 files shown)


jvyden (approved):
Yeah this system never really worked.

This was done for security reasons (theoretically commands could be injected as you're publishing a copied level) but it's fine as we explicitly do not do trusted operations in filtering.

Issue #945, opened 2025-09-08
most LBP1 levels are marked as 1 player only


Arcadius2006:
no clue why this happens

Toastbrot236:
Likely because of lbp1 not setting the min/max player XML elements, making Refresh default to 0 for both, which gets clamped to 1 during the publish. This is already fixed with #935

Toastbrot236:
The fix should be in the latest release now, but we will likely have to add a migration to force already uploaded lbp1 levels to have a maximum player count of 4. Or have those who cant wait simply try to update their lbp1 levels.

jvyden:
I can write a quick migration for this.

PR #955, opened 2025-09-19
Add migration to fix min/max players that are out of range

jvyden:


Refresh.Interfaces.Workers/Migrations/ClampPlayerLimitsMigration.cs
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    protected override void Migrate(WorkContext context, GameLevel[] batch)
+    {
+        foreach (GameLevel level in batch)
+        {
+            if (level.MinPlayers == 0 || level.GameVersion == TokenGame.LittleBigPlanet1)
+                level.MinPlayers = 1;
+
+            if (level.MaxPlayers == 0 || level.GameVersion == TokenGame.LittleBigPlanet1)
```
Toastbrot236, line 23: Should probably check whether `MaxPlayers` is not 4 here, since the problem in #945 was that the publish endpoint used to clamp both `MinPlayers` and `MaxPlayers` to atleast 1, effectively setting both to 1 for all LBP1 levels.

PR #961, opened 2025-10-04
Fix and improve hub challenges

Toastbrot236:
This fixes challenge DB methods to work under EF and Postgres, optimizes them and the challenge game endpoints, gets rid of challenge expiration as that is counter-productive, and reworks many of my comments and summaries. Also adds more slot types to `GameSlotType` and `RoomSlotType` in the process.


Refresh.Core/Services/ChallengeGhostRateLimitService.cs
```diff
@@ -4,6 +4,9 @@
 
 namespace Refresh.Core.Services;
 
+/// <summary>
+/// Documentation on why this exists: <see cref="Endpoints.Game.ChallengeEndpoints.GetContextualScoresForChallenge"/>
+/// </summary>
```
jvyden: Probably best to just move that here instead of keep it in endpoint code.

Refresh.Core/Services/ChallengeGhostRateLimitService.cs
```diff
@@ -14,49 +17,37 @@ public ChallengeGhostRateLimitService(Logger logger, TimeProviderService timePro
...
     {
         // Unconditionally add the user to the set
         this._challengeGhostRateLimitedUsers.Add(id, this._timeProviderService.TimeProvider.Now);
     }
 
-    /// <summary>
-    /// Removes the given user ID from a dictionary containing the IDs of users whose challenge ghost 
-    /// download requests made by LBP Hub should be temporarily blocked
-    /// </summary>
-    /// <param name="id">The user ID</param>
```
jvyden: Why remove xmldoc?

Refresh.Database/GameDatabaseContext.Challenges.cs
```diff
@@ -10,6 +10,10 @@ public partial class GameDatabaseContext // Challenges
 {
     #region Challenges
 
+    private IQueryable<GameChallenge> GameChallengesIncluded => this.GameChallenges
+        .Include(c => c.Level)
+        .Include(c => c.Level.Publisher);
```
jvyden:
```suggestion
    private IQueryable<GameChallenge> GameChallengesIncluded => this.GameChallenges
        .Include(c => c.Level)
        .Include(c => c.Level.Statistics)
        .Include(c => c.Level.Publisher);
```

Refresh.Interfaces.Game/Endpoints/ChallengeEndpoints.cs
```diff
@@ -167,7 +215,7 @@ public Response SubmitChallengeScore(RequestContext context, DataContext dataCon
                 +"Try to submit another score!",
                 user
             );
-            dataContext.Logger.LogDebug(BunkumCategory.UserContent, $"Ghost asset with hash {body.GhostHash} is corrupt");
+            // Reason is already being logged by both the deserialization and the validation methods
             return BadRequest;
```
jvyden:
```suggestion
            );
            
            return BadRequest;
```

Refresh.Interfaces.Game/Types/Challenges/LbpHub/SerializedChallenge.cs
```diff
@@ -70,8 +71,9 @@ public IEnumerable<GameChallengeCriteriaType> CriteriaTypes
             PublisherName = old.Publisher?.Username ?? SystemUsers.DeletedUserName,
             StartCheckpointUid = old.StartCheckpointUid,
             FinishCheckpointUid = old.FinishCheckpointUid,
-            PublishedAt = old.PublishDate.ToUnixTimeMilliseconds(),
-            ExpiresAt = old.ExpirationDate.ToUnixTimeMilliseconds(),
+            PublishedAt = DateTimeOffset.MaxValue.ToUnixTimeMilliseconds(),
```
jvyden: Is there a specific reason to be using MaxValue here for `PublishedAt` here? That seems unrelated.

Toastbrot236: ExpiresAt is at MaxValue because it may not be in the past, else the game will take care of hiding the challenge as expired by itself.

The game doesn't show both timestamps anywhere, it only seems to calculate the difference from now to expiration to show it in the "x days left" span, and PublishedAt doesn't seem to even be used anywhere, but iirc it may not be too small or else the game will crash.

No real reason for PublishedAt to be at MaxValue. I can just make it be set to the real timestamp.

Toastbrot236: As it turns out, setting `PublishedAt` to a timestamp too far behind from now will crash the game, so we should keep it at MaxValue

jvyden:
pls click 'resolve conversation' when fixing things

Toastbrot236:
> pls click 'resolve conversation' when fixing things

Ok I will do that from now on. I've done this rarely until now because i wasn't always sure you're 100% ok with my changes

PR #967, opened 2025-10-14
Include 'clientsConnected' in GameUserResponses

Toastbrot236:
Required by LBP3 to enable these two buttons

image

### DIFF
--- a/Refresh.Interfaces.Game/Types/UserData/SerializedClientsConnected.cs
+++ b/Refresh.Interfaces.Game/Types/UserData/SerializedClientsConnected.cs
@@ -1,0 +1,15 @@
+using System.Xml.Serialization;
+
+namespace Refresh.Interfaces.Game.Types.UserData;
+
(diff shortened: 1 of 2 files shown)


Refresh.Interfaces.Game/Types/UserData/SerializedClientsConnected.cs
```diff
@@ -0,0 +1,15 @@
+using System.Xml.Serialization;
+
+namespace Refresh.Interfaces.Game.Types.UserData;
+
+[XmlRoot("clientsConnected")]
+[XmlType("clientsConnected")]
+public class SerializedClientsConnected
+{
+    // Just set all to true, we don't track this data and probably don't want to either.
```
jvyden: We *will* be tracking this information in the future, to show what games a user owns/plays on a given platform. This is one of the things badges/hidden achievements will be looking for in the future.

jvyden (approved):

PR #974, opened 2025-10-20
Implement detecting and hiding modded planets if preferred

### LINKED ISSUE #968: Option to hide modded planets
I think there should be an opt-in option, similar to showMods, which users can set to hide modded planets which are not theirs. Whether a planet for a game is modded would be stored on the GameUser, similarly to how whether a level is modded is stored on the GameLevel.

This is useful for browsing levels in-game, especially if the user's internet is not very good or they're playing lbp3, which periodically forces you to wait until the publisher's planet has finished loading while scrolling levels. Loading can take a really long time depending on how many modded assets the planet has and how large they are. Can also be annoying in other games (when actually selecting a level to play from the list, which requires loading the planet), but not as much as in lbp3.

Detecting modded planets would also be a little stricter. If a planet level depends on custom Materials or especially VoiceRecordings, it should already be flagged as modded. These are not only impossible to put on a planet without modding, but detecting the latter asset type would also make this feature useful for those who want to scroll levels without being randomly jumpscared by low quality recordings.

I think this option should be its own toggle, independent from showMods, because i can imagine enough users (myself included) would want to hide modded planets, but not modded levels.

Toastbrot236:
Closes #968

### DIFF
--- a/Refresh.Core/Refresh.Core.csproj
+++ b/Refresh.Core/Refresh.Core.csproj
@@ -22,7 +22,7 @@
         <PackageReference Include="CommandLineParser" Version="2.9.1" />
         <PackageReference Include="Discord.Net.Webhook" Version="3.18.0" />
         <PackageReference Include="DnsClient" Version="1.8.0" />
-        <PackageReference Include="MailKit" Version="4.14.0" />
+        <PackageReference Include="MailKit" Version="4.14.1" />
         <PackageReference Include="NPTicket" Version="3.1.1" />
         <PackageReference Include="Pfim" Version="0.11.3" />
         <PackageReference Include="SharpZipLib" Version="1.4.2" />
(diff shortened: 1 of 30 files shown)


Refresh.Database/Extensions/GameDatabaseContextExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        GameAsset? rootAsset = database.GetAssetFromHash(rootAssetHash);
+        rootAsset?.TraverseDependenciesRecursively(database, (_, asset) =>
+        {
+            if (asset != null && (asset.AssetFlags & (AssetFlags.Modded | AssetFlags.ModdedOnPlanets)) != 0)
+                modded = true;
+        });
+        
+        return modded;
+    }
```
jvyden: These definitely don't belong as an extension.

PR #995, opened 2025-12-27
Allow resetting user avatars

Toastbrot236:
This allows users to reset their own avatar/profile icon both in-game and through the API. Doing so makes the console games show the user's PSN avatar, which is a [requested feature](https://discord.com/channels/1049223665243389953/1049223665998372947/1454213330062278870).

If a user tries resetting their icon in-game, the user update request will have the icon XML element set to an empty value.


Toastbrot236:
Maybe adding some tests for this might be good, even though I've already tested this manually...

PR #1043, opened 2026-03-08
Stricter rate-limits for many endpoints

Toastbrot236:
This PR assigns "unique" rate-limit settings to many game and API endpoints, making certain endpoints have stricter limits than others. Most of these settings are stored in their own classes so multiple endpoint classes could share them, and to organize them more clearly. Not all rate-limit settings are in their own classes yet. The global rate-limit is also slightly stricter for now.

Not all settings are properly tested yet. Some may, in reality, be too strict or too loose; they are experimental.


Toastbrot236:
Seems to work fine for the mainline games and refresh-web...

PR #1046, opened 2026-03-09
Split GamePhotoSubjects into their own DB table

Toastbrot236:
Makes working on photos easier and less chaotic at the cost of one more DB call per requested photo. This is not that bad, because I do already have something in mind (and partially implemented) which will mostly avoid that and other DB calls when requesting photos, but I have decided to PR that separately, so this PR is more of a preparation for photo optimization.


Toastbrot236:
The latest commit removes that extra DB call per serialized photo, so now photo fetching is as fast as previously, while still keeping subjects in their own table internally.

PR #1058, opened 2026-03-21
Block more endpoints if read-only mode is enabled

Toastbrot236:
This makes a few adjustments to read-only mode:
- Now all upload/modification endpoints for both the game and public API are blocked if the user has no perms due to this mode (still excluding account/profile updating, contest updating and moderation)
- The only 2 deletion endpoints which were previously blocked (playlist deletion) are no longer affected by read-only mode. This is mostly for consistency, and because this mode is primarily intended to block new content.
- The user will now be properly notified if their level publish has failed because of this.


FeTetra (approved):
lgtm

PR #1066, opened 2026-04-01
Properly fix website pin migration

Toastbrot236:
Found another pagination issue with this migration right after merging the previous PR, so I've (partially) reworked it to iterate users instead of pin relations.

This does make the newly introduced feature to alter the processed and total counts in the `JobState` unused for now, but I have not removed it because it might eventually become necessary, if a different job ever needs to delete entities.


Toastbrot236:
Tested this once more, it works

PR #1080, opened 2026-04-27
Add configurable upload rate-limits for photos and playlists

Toastbrot236:
Does various renames and reworks to the configurable level upload rate-limit feature to make it easily usable on other kinds of entities aswell, and adds such rate-limits for photos and playlists. Also, user-related rate-limit data stored in DB is now stored in a separate table instead of `GameUsers`.

Needs a few more tests before this can be merged, and while I have managed to fix the level rate-limit tests which were being skipped before, they have broken again now, so they need to also be fixed again before the merge.


Toastbrot236:
+300 added lines just from tests alone...

FeTetra (approved):

PR #1086, opened 2026-05-29
Implement helper method for asset reference validation

Toastbrot236:
This PR implements a helper method which applies various common validations to a given asset reference (common as in these kinds of validations are needed by various endpoints, like e.g. multiple endpoints needing to ensure a referenced asset is an image, or not a GUID, or not disallowed etc). The idea is that, instead of every endpoint implementing its own validation, they'd just call this method and handle its return value.

This would deduplicate a relatively high amount of code, and it would potentially also prevent edge cases where some endpoints forget to validate some things (e.g. only root level hashes are ensured to be valid hashes right now, some API endpoints only ensure an asset is in the database and not the data store, the photo upload endpoint only ensures a photo's dependencies are in the data store and not the database etc).

Since a complete refactor would've made this PR way larger than it already is (~850 lines just to implement and test the method), this PR only implements and tests the method. Modifying endpoint methods to actually use this method would have to be done in future PRs.

### DIFF
--- a/Refresh.Core/Helpers/ResourceValidationHelper.cs
+++ b/Refresh.Core/Helpers/ResourceValidationHelper.cs
@@ -1,0 +1,116 @@
+using System.Diagnostics;
+using Bunkum.Core;
+using NotEnoughLogs;
+using Refresh.Common.Verification;
+using Refresh.Core.Types.Assets.Validation;
+using Refresh.Database.Models.Assets;
+using Refresh.Database.Models.Authentication;
+
+namespace Refresh.Core.Helpers;
+
+public abstract class ResourceValidationHelper
+{
+    public static ValidatedAssetResult Validate(AssetValidationParameters parameters, Logger logger)
+    {
+        string assetTypeStr = parameters.AssetContextTypeStr ?? (parameters.MustBeTexture ? "image asset" : "asset");
+        GameAsset? asset = null;
+        bool existsInDataStore = false;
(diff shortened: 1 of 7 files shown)


Refresh.Core/Helpers/ResourceValidationHelper.cs
```diff
@@ -0,0 +1,120 @@
...
+                    sw.Stop();
+                    logger.LogDebug(BunkumCategory.UserContent, $"Failed to get '{filename}' after {sw.ElapsedMilliseconds}ms.");
+                    return new(BadRequest, null, $"The used {assetTypeStr} was invalid or corrupt.", onNewAssetRefCallback, existsInDataStore: existsInDataStore);
+                }
+
+                sw.Stop();
+                logger.LogInfo(BunkumCategory.UserContent, $"Successfully imported '{filename}' in {sw.ElapsedMilliseconds}ms.");
+            }
+
+            // FIXME: for some reason, PSP texture detection/conversion broke so we can no longer tell if a PSP texture is actually a texture, so skip this for PSP
```
FeTetra, line 106: Do you know when?

Toastbrot236: No, just noticed it randomly on production levels

FeTetra (approved):

Refresh.Core/Helpers/ResourceValidationHelper.cs
FeTetra: I see, not a big deal for now i guess.

PR #1100, opened 2026-07-09
Ability to see previous usernames via API

Toastbrot236:
Adds API endpoints which allow any user to see their own previous usernames, and also allow mods/admins to get all users which have ever used a certain name, aswell as all usernames previously used by any user.

Needs tests, and probably also the ability to look up users by previous name case-insensitively.


Toastbrot236:
Removed one of the 3 endpoints because of some deeper issues with Bunkum, but now this only needs a few unit tests and it's done.

FeTetra (approved):

PR #1116, opened 2026-08-22
Add a configurable endpoint rate-limiter

Toastbrot236:
This ports Bunkum's `RateLimiter` and modifies it in a way where, instead of getting its bucket data from hard-coded constants, it instead gets the data from a config file named `endpointRateLimits.json`. This PR only implements and generally tests this rate-limiter, and only makes 3 endpoints use this new rate-limiter (for testing purposes), however, after this PR, I'd like to do follow-up PRs where I make all our endpoints use this new rate-limiter.

There are mainly 2 reasons why we want configurable rate-limits:
- If our limits turn out to be too strict or too lenient somewhere again, instead of having to open a new PR and push a new release every time we want to change the constants, we would now only have to change the wanted values in the config and restart the server
- Some instances might be able to handle more requests than others depending on their infrastructure, so instead of us having to decide constants for everyone, server owners would now be able to choose themselves how strict their rate-limits should be

Also, I'm not exactly sure how we'd want to handle migrations in the future where we want to update the default values of existing buckets. Should we just keep this one `OverwriteBucketValuesIfDefaultsAreUpdated` config flag, or should we have such a flag for every bucket separately?


FeTetra (approved):

FeTetra:
I saw the public bucket storage record that was only like 3 lines in its own file, I would've asked about it but I assume it just didn't fit anywhere else.

Toastbrot236:
> I saw the public bucket storage record that was only like 3 lines in its own file, I would've asked about it but I assume it just didn't fit anywhere else.

I thought to put the record into an extra file like with `ScoreWithRank` and `ChallengeScoreWithRank`.

